PR #11, opened 2025-04-11
ruff check . --fix

Cascoopman:
Running ruff reveals and solves some unused imports and unnecessary f strings.


turanbulmus (approved):
Thanks a lot for the ruff check

PR #19, opened 2025-04-13
Fixed Required packages for FOMC Agent example

MG-Cafe:
There were 2 more packages missing to run this agent and they were not mentioned in the readme file, so added there

### DIFF
--- a/agents/fomc-research/README.md
+++ b/agents/fomc-research/README.md
@@ -94,7 +94,7 @@ to implement this workflow.
 
     Install the FOMC Research agent requirements:
     ```bash
-    poetry install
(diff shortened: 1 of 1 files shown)


mstyer-google (commented):
Poetry should install all the requirements. Are you finding that doesn't work?

anifort:
Hello @MG-Cafe , We had a look at the suggested change and we would not be applying this to the base code.
`poetry install` is the right command to install all necessary dependencies. However it seems that you did not have poetry already install the first time you run the script. We are going to make the instructions a bit more clear to read:

If you have not installed poetry before run `pip install poetry` first before using `poetry install

I hope this clears out this confusion

PR #26, opened 2025-04-14
Fix hyperlinks in customer-service README

kevinam99:


turanbulmus (approved):
thanks for the corrections Kevin!

PR #31, opened 2025-04-14
Add security guardrails on tools in custumer_service sample agent

mmontan:
More information can be found in the https://google.github.io/adk-docs/guides/responsible-agents/

The guardrails aim at demonstrating tool-limitations as a mechanism to reduce the risk of malicious prompts exploiting the agent to access data about other users.


anifort (approved):
LGTM

PR #86, opened 2025-05-12
Fix: Use Python -m flag in deployment script to resolve module import

duboc:
When running the deployment script directly, Python couldn't find the brand\_search\_optimization module. Using the -m flag ensures Python treats the script as a module and adds the current directory to the path.


sw-savop:
I am having a similar issue when attempting to deploy the Customer Service example. For me it fails to find the 'customer_service' module. 

I've tried the -m solution but to no avail. Any tips on how to fix?

turanbulmus (approved):

turanbulmus:
thanks a lot for raising this. It is approved @sw-savop for your concern, can you file an issue so that we can take a look?

Issue #27, opened 2025-04-14
Add google-adk[extensions] to some samples deps


Jacksunwei:
E.g. for rag, it uses retrieval package, which will require installing `google-adk[extensions]` for LlamaIndex.

Context: https://github.com/google/adk-python/issues/96

turanbulmus:
@cnemri can you take a look at this?

PR #89, opened 2025-05-15
Fix: added extensions to google adk deps for RAG sample agent

cnemri:
Fixes this issue https://github.com/google/adk-samples/issues/27


mstyer-google (approved):
LGTM

PR #90, opened 2025-05-15
image scoring agent changes

kishorerj:
I have created an agentic app to generate images for news articles and score the same against defined policies.. Please take a look and check if this can be part of Agent Garden..Also please provide your high level review comments. I will improve the prompts and create tests etc once I hear back. U can currently run this via 'adk web'


kishorerj:
@mstyer-google @koverholt

PR #148, opened 2025-05-31
fix: Use final static BaseAgent ROOT_AGENT field.

vorburger:
In line with https://github.com/google/adk-docs/pull/350.


ivanmkc (approved):
LGTM

PR #250, opened 2025-07-09
Enable separate data and compute projects for BQ

mstyer-google:


python/agents/data-science/data_science/sub_agents/bigquery/tools.py
```diff
@@ -28,11 +28,12 @@
...
-# Assume that `BQ_PROJECT_ID` is set in the environment. See the
-# `data_agent` README for more details.
-project = os.getenv("BQ_PROJECT_ID", None)
+# Assume that `BQ_COMPUTE_PROJECT_ID` and `BQ_DATA_PROJECT_ID` are set in the
+# environment. See the `data_agent` README for more details.
+data_project = os.getenv("BQ_DATA_PROJECT_ID", None)
+compute_project = os.getenv("BQ_COMPUTE_PROJECT_ID", None)
 location = os.getenv("GOOGLE_CLOUD_LOCATION", "us-central1")
-llm_client = Client(vertexai=True, project=project, location=location)
+llm_client = Client(vertexai=True, project=compute_project, location=location)
```
slilichenko: Vertex AI project and BQ compute projects are not the same. I would add another configuration for that.

mstyer-google: Thanks Sergei, good catch - this should be GOOGLE_CLOUD_PROJECT instead, not either one of the BQ projects.

python/agents/data-science/data_science/sub_agents/bigquery/chase_sql/qp_prompt_template.py
```diff
@@ -75,14 +75,14 @@
 
 **************************
 【Table creation statements】
-CREATE TABLE {BQ_PROJECT_ID}.restaurant.generalinfo
+CREATE TABLE {BQ_DATA_PROJECT_ID}.restaurant.generalinfo
```
slilichenko: To make things bullet-proof, the project id need to be enclosed in backticks. This would prevent failures for project ids which include characters which cause problems to the SQL parser.

mstyer-google: Good catch here as well.

5Y5TEM (approved):
LGTM

PR #279, opened 2025-07-24
bugfix(camel): fix incorrect reader type check in CaMeL demo

hironow:
This fixes a bug where reader != readers.Public() raised an error due to reader being a tuple.
Now correctly compares the first element: reader[0] != readers.Public().

This change ensures compatibility with the current output format of get_all_readers, which returns a tuple of (Readers[Any], frozenset[int]).


hironow:
@5Y5TEM Please take a look at this PR when you have time. Feedback is welcome!

PR #298, opened 2025-08-07
feat: use ADK built-in BigQuery tools in the data science agent

shobsi:
This change cimplifies the custom tools implementation for BigQuery interactions.
Ref. https://google.github.io/adk-docs/tools/built-in-tools/#bigquery

### DIFF
--- a/java/agents/software-bug-assistant/software-bug-assistant/src/main/java/SoftwareBugAssistant.java
+++ b/java/agents/software-bug-assistant/software-bug-assistant/src/main/java/SoftwareBugAssistant.java
@@ -32,8 +32,8 @@ public class SoftwareBugAssistant {
     private static String NAME = "SoftwareBugAssistant";
     private static String USER_ID = "test-user";
 
-    // ROOT_AGENT needed for ADK Web UI. 
-    public static BaseAgent ROOT_AGENT = initAgent();
+    // ROOT_AGENT needed for ADK Web UI.
+    public static final BaseAgent ROOT_AGENT = initAgent();
 
 
     public static void main(String[] args) {
@@ -65,7 +65,7 @@ public class SoftwareBugAssistant {
 
 
     public static BaseAgent initAgent() {
-        // Set up MCP Toolbox connection to Cloud SQL 
+        // Set up MCP Toolbox connection to Cloud SQL
         try {
             String mcpServerUrl = System.getenv("MCP_TOOLBOX_URL");
             if (mcpServerUrl == null || mcpServerUrl.isEmpty()) {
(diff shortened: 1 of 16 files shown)


python/agents/data-science/data_science/sub_agents/bigquery/agent.py
```diff
@@ -35,6 +41,30 @@ def setup_before_agent_call(callback_context: CallbackContext) -> None:
             tools.get_database_settings()
 
 
+def handle_after_tool_call(
```
mstyer-google: Could you give this a more descriptive name? Something like 'store_results_in_context'. That way when someone looks at the agent definition it will be more clear what's happening in the `after_tool_callback`. In general I think that's a pattern we should promote.

python/agents/data-science/data_science/sub_agents/bigquery/prompts.py
```diff
@@ -30,7 +30,28 @@ def return_instructions_bigquery() -> str:
         db_tool_name = None
         raise ValueError(f"Unknown NL2SQL method: {NL2SQL_METHOD}")
 
-    instruction_prompt_bqml_v1 = f"""
+    instruction_prompt_bq_v2 = f"""
```
mstyer-google: Please call this variable `instruction_prompt_bigquery` -- that will be one less merge conflict I have to deal with when I merge my branch.

python/agents/data-science/data_science/sub_agents/bigquery/prompts.py
```diff
@@ -30,7 +30,28 @@ def return_instructions_bigquery() -> str:
...
+          "nl_results": "Natural language about results, otherwise it's None if generated SQL is invalid"
+      
+      You should pass one tool call to another tool call as needed!
+
+      NOTE: you should ALWAYS USE THE TOOL {db_tool_name} to generate SQL, not make up SQL WITHOUT CALLING TOOLS.
+      Keep in mind that you are an orchestration agent, not a SQL expert, so use the tools to help you generate SQL, but do not make up SQL.
+
+    """
+
+    instruction_prompt_bq_v1 = f"""
```
mstyer-google: You can also get rid of the other prompt versions.

python/agents/data-science/data_science/sub_agents/bigquery/agent.py
shobsi: called it `store_results_in_context`, thanks!

python/agents/data-science/data_science/sub_agents/bigquery/prompts.py
shobsi: done, thanks!

python/agents/data-science/data_science/sub_agents/bigquery/prompts.py
shobsi: I guess I was just following the convention in https://github.com/google/adk-samples/blob/main/python/agents/data-science/data_science/sub_agents/bqml/prompts.py to keep older prompts. Removed the older prompt from here. Do you want me to do the same in bqml agent too?

mstyer-google: We should do that too but we can do it in a separate PR.

PR #301, opened 2025-08-07
adding trends and insights agent

jswortz:
Adding an agent we have been using with retail and marketing to highlight ADK + Multimedia use cases. This also comes with tooling to deploy on agentspace. @tottenjordan


happyhuman:

PR #303, opened 2025-08-08
Feat tf data science

eliasecchig:


python/agents/data-science/README.md
```diff
@@ -53,121 +53,51 @@ The key features of the Data Science Multi-Agent include:
     cd adk-samples/python/agents/data-science
     ```
 
-2.  **Install Dependencies with Poetry:**
+2.  **Install Dependencies with uv:**
```
mstyer-google: The data science agent was updated to use uv in a recent PR. If you sync your fork, you'll get those changes and you won't need to include them here.

python/agents/data-science/README.md
```diff
@@ -53,121 +53,51 @@ The key features of the Data Science Multi-Agent include:
...
+5.  **Environment Variables (Optional):**
+    
+    The agent automatically detects your Google Cloud project and uses sensible defaults for all configuration. If you're using the default project setup, no additional environment variables are needed.
+    
+    If you need to customize the configuration, you can create a `.env` file in the project root directory with any of the variables output by Terraform.
 
+    For our sample use case, we are working on the Forecasting Sticker Sales data from Kaggle:
     _Walter Reade and Elizabeth Park. Forecasting Sticker Sales. https://kaggle.com/competitions/playground-series-s5e1, 2025. Kaggle._
 
-    *   First, set the BigQuery project IDs in the `.env` file. This can be the
```
mstyer-google, line 106: It looks like you're removing the rest of the setup steps -- I don't think we want to force users to use the Agent Starter Pack, just make it an option. Can we keep these instructions, but also give instructions for setting it up using Terraform and the ASP?

python/agents/data-science/data_science/config.py
mstyer-google, line 1: Adding this config system looks like a good change in itself, separate from the deployment method. It's also a pretty far-reaching change to make alongside another big change like adding terraform. Could you pull this out into a separate PR? That way we can review and merge each big change separately.

python/agents/data-science/README.md
```diff
@@ -267,114 +197,52 @@ Tests assess the overall executability of the agents.
...
-The last sequence of digits is the AgentEngine resource ID.
+This command handles creating the project without needing to pre-install the package into a virtual environment.
+</details>
 
-Once you have successfully deployed your agent, you can interact with it
-using the `test_deployment.py` script in the `deployment` directory. Store the
-agent's resource ID in an environment variable and run the following command:
+You'll be prompted to select a deployment option (Agent Engine or Cloud Run) and verify your Google Cloud credentials.
 
+### Step 2: Deploy to Development Environment
```
mstyer-google: I think we want to include instructions for how to deploy this using the ADK command line, e.g. `adk deploy agent_engine` and `adk deploy cloud_run`, not just Terraform. I saw you deleted all the deployment code elsewhere -- that's fine, as long as it's really not necessary and we include instructions for running the command line deployment.

python/agents/data-science/deployment/test_deployment.py
mstyer-google, line 1: This is a useful script for folks wanting to test the deployment immediately after they've run it. I suggest we keep it, but just make it work with whatever deployment method we're using.

python/agents/data-science/infra/README.md
```diff
@@ -0,0 +1,307 @@
...
+
+3. **Build and Deploy Agent:**
+   ```bash
+   # Build Python wheel
+   cd ..
+   uv build --wheel --out-dir deployment
+   
+   # Deploy to Agent Engine
+   cd deployment/
+   uv run python deploy.py --create
```
mstyer-google: It looks like this file got deleted -- does this command still work?

python/agents/data-science/README.md
zeroasterisk: There are really 2 approaches today: Agent Starter Pack (terraform) and ADK deploy commands.

I'm a big fan of ASP as is Julia.  

Perhaps we have the "easy via ASP" option on top and "manual via commands" below?  this is a common approach.

Issue #292, opened 2025-08-03
Does the MLE-STAR project need to have Torch installed ???


JuanVargas:
I am trying to work with the Machine Learning Enginering project. (MLE-STAR).  

The project looks quite involved. According to the chart in the README.md there are about 22+ different agents.

This is more of a question than a bug.   When I ran 

``` poetry install ```

I was surprised to see that the instlal did not finish because Pytorch is NOT installed on my virt env. (I have it in other envs, but since the documentation of the project does not mention Torch, I did no tinstall it.

If having Torch is required to run this project, why is that not mentioned in the setup in README.md  ???

If there is a way to run the project without having Torch, please let us know.

Thank you

axiom-of-choice:
+1
Seems to be a bug, because i followed all the instructions and when executing `poetry run adk run machine_learning_engineering` i got the error of torch not installed.

JuanVargas:
Yes, that must be a bug. I removed the line in Poetry where Torch is declared as q pre-req, but that took me nowhere. Only after installing Torch via Pip under my env where the ADK was running, I was able to get the code work. But then, after that, the  second issue I faced was that the code takes too many resources to execute, so much, that in my case, at the middle of the run, I got a message telling me that my resources had been exhausted, and the application aborted.

PR #311, opened 2025-08-12
fix: Correct PyTorch installation in pyproject.toml

ivanmkc:
This commit fixes the PyTorch installation by using a dedicated source instead of a direct URL. This makes the installation platform-agnostic and resolves the issue where the previous URL was specific to Linux.

Fixes https://github.com/google/adk-samples/issues/292


ivanmkc:
PTAL: @happyhuman

ivanmkc:
PTAL @raj-sinha

happyhuman (approved):

Issue #88, opened 2025-05-14
Cannot create a new chat session with a RAG agent using corpus


dinh-vu-1:
I am following the guides on using agent with RAG here: https://github.com/google/adk-samples/blob/main/agents/RAG/README.md

Setup and done things in order, When I create and deploy an agent without giving it access to RAG, it does work:

```
root_agent = Agent(
    model='gemini-2.0-flash-001',
    name='ask_rag_agent_007',
    instruction='You are a helpful agent that reply in a happy tone.',
)
```

However, when deploy an agent (doing all the steps per README file) with tools to use RAG Corpus, it's returning error when trying to create a chat session:

Traceback (most recent call last):
  File "/Users/dinh.vu/Library/Caches/pypoetry/virtualenvs/rag-IE5NJtFt-py3.11/lib/python3.11/site-packages/google/api_core/grpc_helpers.py", line 76, in error_remapped_callable
    return callable_(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dinh.vu/Library/Caches/pypoetry/virtualenvs/rag-IE5NJtFt-py3.11/lib/python3.11/site-packages/grpc/_interceptor.py", line 277, in __call__
    response, ignored_call = self._with_call(
                             ^^^^^^^^^^^^^^^^
  File "/Users/dinh.vu/Library/Caches/pypoetry/virtualenvs/rag-IE5NJtFt-py3.11/lib/python3.11/site-packages/grpc/_interceptor.py", line 332, in _with_call
    return call.result(), call
           ^^^^^^^^^^^^^
  File "/Users/dinh.vu/Library/Caches/pypoetry/virtualenvs/rag-IE5NJtFt-py3.11/lib/python3.11/site-packages/grpc/_channel.py", line 440, in result
    raise self
  File "/Users/dinh.vu/Library/Caches/pypoetry/virtualenvs/rag-IE5NJtFt-py3.11/lib/python3.11/site-packages/grpc/_interceptor.py", line 315, in continuation
    response, call = self._thunk(new_method).with_call(
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dinh.vu/Library/Caches/pypoetry/virtualenvs/rag-IE5NJtFt-py3.11/lib/python3.11/site-packages/grpc/_channel.py", line 1198, in with_call
    return _end_unary_response_blocking(state, call, True, None)
[...]

PR #317, opened 2025-08-12
fix: rag agent fix, updated google-adk and google-cloud-aiplatform versions

cnemri:
Fixes issue #88 
Also, please make sure to run ./deployment/grant_permissions.sh to grant your agent engine agent access to your rag engine.


ivanmkc (approved):
LGTM

PR #321, opened 2025-08-15
Removed keyring instruction from README.md in data-science

mstyer-google:
The data-science agent was updated to use uv for package management instead of uv, so the instruction regarding updating the keyring during poetry installation is out of date.

### DIFF
--- a/python/agents/data-science/README.md
+++ b/python/agents/data-science/README.md
@@ -55,12 +55,6 @@ The key features of the Data Science Multi-Agent include:
 
 2.  **Install Dependencies with uv:**
 
-    **Note for Linux users:** If you get an error related to `keyring` during the installation, you can disable it by running the following command:
-    ```bash
-    poetry config keyring.enabled false
-    ```
-    This is a one-time setup.
-
     ```bash
     uv sync
     ```


python/agents/data-science/README.md
```diff
@@ -55,25 +55,30 @@ The key features of the Data Science Multi-Agent include:
...
+    virtual environment will be created in a `.venv` directory inside
+    `adk-samples/python/agents/data-science`. If you already have a virtual
+    environment created, or you want to use a different location, you can use
+    the `--active` flag for `uv` commands, and/or change the
+    `UV_PROJECT_ENVIRONMENT` environment variable. See
+    [How to customize uv's virtual environment location](https://pydevtools.com/handbook/how-to/how-to-customize-uvs-virtual-environment-location/)
+    for more details.
 
-3.  **Activate the uv Shell:**
+2.  **Activate the uv Shell:**
```
jackwotherspoon, line 73: this step is not required with uv...right?

mstyer-google: Not super familar with `uv` myself, but I believe the way it works is that if you run everything using `uv run ...` consistently, you don't need to activate the environment, but if you want to be able to run scripts etc. normally, without using `uv run`, then you do. My feeling is it's better to prompt users to activate the environment so when they try running commands like `adk web`, the correct environment is active.

(Basing this on the documentation [here](https://docs.astral.sh/uv/guides/projects/#running-commands)).

jackwotherspoon: SGTM, I am fine having the extra step if we think it will cause less errors for users 👍

PR #357, opened 2025-08-29
Add two safety security guardrails into ADK samples: Gemini as a Judge and Model Armor as a Plugin

sammalik-lgtm:
This repository provides an example of a multi-agent system built with the Agent Development Kit (ADK), focusing on how to implement global safety guardrails using ADK's plugin feature. The system includes two distinct safety plugins: one that uses an agent as a judge and another that leverages the Model Armor API.

It adds LLMAsAJudge, and ModelArmor as a safety filter


python/agents/safety-plugins/pyproject.toml
```diff
@@ -0,0 +1,30 @@
+[tool.poetry]
+name = "safety"
+version = "0.1.0"
+license = "Apache License 2.0"
+description = "Safety Plugins for ADK"
+authors = ["sammalik-lgtm <sammalik@google.com>"]
+readme = "README.md"
```
Sita04, line 7: Is this an optional block that can be removed?

python/agents/safety-plugins/safety_plugins/main.py
```diff
@@ -0,0 +1,99 @@
...
+    runner = InMemoryRunner(
+        agent=root_agent,
+        app_name=APP_NAME,
+        plugins=[
+            # LlmAsAJudge(),
+            # ModelArmorSafetyFilter(
+            #     project_id="YOUR_PROJECT_ID_HERE",
+            #     location_id="us-central1",
+            #     template_id="YOUR_TEMPLATE_ID_HERE",
+            # ),
```
Sita04: Why is this commented out? If one of these has to be enabled, could you please add a comment here.

python/agents/safety-plugins/safety_plugins/main.py
```diff
@@ -0,0 +1,99 @@
...
+from . import prompts
+from . import util
+
+
+Agent = llm_agent.LlmAgent
+LlmAsAJudge = agent_as_a_judge.LlmAsAJudge
+ModelArmorSafetyFilter = model_armor.ModelArmorSafetyFilterPlugin
+InMemoryRunner = runners.InMemoryRunner
+
+
```
Sita04, line 39: Please add a comment here that says not to use hardcoded credentials in the code, but instead use a secret management tool.

python/agents/safety-plugins/safety_plugins/main.py
sammalik-lgtm: Done.

python/agents/safety-plugins/safety_plugins/main.py
sammalik-lgtm: Move everything to .env and secret manager. Updated README

PR #372, opened 2025-09-10
feat: new agent demonstrating price prediction for each step of agentic workflow.

Mukamik:
Created the pricing-predictor agent sample, that demonstrates the use of after_model and after_agent callbacks to predict the approximate one-run, thousand-run and million-run cost of your agentic workflow.

The same callback can be added to any agent to as a demonstration of how to pass LLM-Response data such as token usage between steps in a workflow.

Warning: Pricing predicted or estimated through this tool is not a guarantee of actual billing costs, but a broad, back-of-the-envelope estimation of price of a workflow.


happyhuman:

PR #617, opened 2025-11-10
RAG: Remove hard coded Location from grant permissions and docs

domribaut:
In the RAG python sample, US-central1 is hard-coded in the grant permission script and as default region.
Deploy would fail.
This PR uses the env variable instead and updates the doc  to be consistant with other usage of env variables


happyhuman:
Closing: the recipe this PR targets (python/agents/RAG) has been removed from the repository.

PR #619, opened 2025-11-11
feat: adding bidi-demo sample

kazunori279:
Adding bidi-demo sample which will be the demo application for upcoming ADK Bidi-streaming development guide.


python/agents/bidi-demo/app/main.py
```diff
@@ -0,0 +1,213 @@
...
+# Phase 1: Application Initialization (once at startup)
+# ========================================
+
+app = FastAPI()
+
+# Mount static files
+app.mount("/static", StaticFiles(directory=Path(__file__).parent / "static"), name="static")
+
+# Define your session service
+session_service = InMemorySessionService()
```
hangfei, line 47: maybe for next PRs: could we offer a few options for DBSessionService or VertexSessionService?

hangfei (approved):

PR #621, opened 2025-11-11
chore: add auto-merge workflow for Dependabot PRs

eliasecchig:
Implements automatic approval and merging of Dependabot pull requests once all CI checks pass successfully.

- Auto-approves Dependabot PRs when opened
- Auto-merges Dependabot PRs only after all check suites complete successfully
- Uses squash merge strategy for clean commit history


tpryan (commented):
One worry I have about this is that updating dependencies may result in breakage of the samples which would lead to us having to track down sample owners.  

Any thoughts?

eliasecchig:
@tpryan It's a fair comment. This would auto-merge only if the all tests are passing. Let me restrict this to only samples that are implementing automated tests

PR #645, opened 2025-11-19
feat: add ADK agent to be used in Tau2 bench

selcukgun:
Provides a baseline ADK agent integration that can be registered in tau2 bench.


selcukgun:
Ran the python-checks.sh. It should be good for review now.

happyhuman (approved):

PR #677, opened 2025-11-20
feat: Adding a story teller agent

happyhuman:


python/agents/story_teller/pyproject.toml
tommywagz, line 1: The toml will have to change to comply with the root and the rest of the samples. Right now the hook won't work with this setup

PR #706, opened 2025-11-26
fix: lint bidi-demo with black, isort, and flake8

kazunori279:
fix: lint bidi-demo with black, isort, and flake8

@happyhuman I tried lints with those tools at my local (to avoid bothering you for the future PRs), and got this result. Are these files acceptable for adk-samples repo? Do I need to change settings/configs for those lint tools?


python/agents/bidi-demo/pyproject.toml
```diff
@@ -20,6 +20,9 @@ dependencies = [
 dev = [
     "pytest>=8.0.0",
     "pytest-asyncio>=0.24.0",
+    "black>=24.0.0",
```
happyhuman: There is now a top [pyproject.toml](https://github.com/google/adk-samples/blob/main/pyproject.toml) in the ROOT directory which has some black and isort related setting. I think we should move anything related to formatting to that pyproject.toml to avoid redundant code for each individual sample.

python/agents/bidi-demo/.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
```
happyhuman: There is currently a .flake8 under the ROOT directory. We can just use that.

python/agents/bidi-demo/pyproject.toml
kazunori279: got it, removed lint tool settings.

python/agents/bidi-demo/.flake8
kazunori279: got it, removing.

happyhuman (approved):

PR #704, opened 2025-11-25
chore: Reformatted the brand-search-optimization sample by ensuring black, isort, and flake8 pass.

happyhuman:

PR #725, opened 2025-12-02
chore: Reformatted the data-science sample by ensuring black, isort, and flake8 pass 

happyhuman:


mstyer-google (changes requested):
Same as on #704 - check the line length on the lines you un-indented.

Issue #694, opened 2025-11-24
[BUG]: The Bidi-demo show connecting forever


nghia95:
The Bidi-demo just show "Connecting" forever with the following message: INFO:     
127.0.0.1:46456 - "GET /favicon.ico HTTP/1.1" 404 Not Found

happyhuman:
Closing this issue because the recipe it references (bidi-demo) has been removed from the repository. If you believe this is still relevant to an existing recipe, please open a new issue referencing the current recipe name.

PR #731, opened 2025-12-04
fix issues #694: add connection timeout and improve error handling for bidi-demo WebSocket

wnqqnw19:
```
Fixed the issue where bidi-demo would show "Connecting..." forever when the WebSocket connection failed or timed out.

Changes:
- Added 10-second connection timeout to detect failed connections
- Improved error handling in onerror handler with clearer error messages
- Enhanced onclose handler to properly handle different close codes
- Added timeout cleanup to prevent memory leaks
- Added user-friendly error messages when connection fails
- Prevented unnecessary reconnection attempts when connection times out

This ensures users get proper feedback when the server is not running or connection fails, instead of being stuck on "Connecting..." indefinitely.
```


happyhuman:

PR #786, opened 2025-12-15
Supervised Fine Tuning - Starter Pack Agentic AI Solution using ADK

suddhasatwabhaumik:
As enterprise needs grow more sophisticated, standard prompting techniques are proving insufficient for complex scenarios. This has elevated the importance of fine-tuned Large Language Models (LLMs) for specialized tasks like nuanced Text-to-SQL generation and advanced conversational systems. However, the iterative and resource-intensive nature of fine-tuning presents a significant bottleneck to rapid deployment.

To address this, we propose the creation of an Autonomous Fine-Tuning Agent, built using the Agent Development Kit (ADK) on Google Cloud. This agent reimagines the model customization workflow by moving from a manual, step-by-step process to a goal-oriented, autonomous system. Instead of a user-driven tool, this agent leverages an AI-native approach to manage the entire fine-tuning lifecycle. The core of this agent is a reasoning engine, like Gemini, that plans and executes a series of tasks to achieve a desired outcome.

The process begins with an AI engineer or client providing the agent with a high-level objective, such as achieving a target accuracy for a specific task. The user equips the agent with a small seed set of 20-50 examples and a ground truth dataset for evaluation. From there, the agent operates autonomously, using a suite of specialized tools:

- Synthetic Data Generation: The agent intelligently analyses the seed data and determines the optimal strategy for generating a high-quality, scaled-up dataset of thousands of examples.
- Supervised Fine-Tuning (SFT): It programmatically initiates the fine-tuning job on a user-selected base model, such as Gemini Flash.
- Automated Evaluation & Iteration: Upon completion, the agent evaluates the new model against the ground truth dataset. Critically, it then analyzes the results. If the desired accuracy is not met, the agent autonomously decides on the next steps—whether to generate more diverse synthetic data, refine the existing set, or attempt another fine-tuning run. This iterative loop continues until the agent's goal is achieved or human intervention is requested.

By packaging this intelligence into an agent built on the ADK on Google Cloud, we empower our teams and clients to delegate the complex engineering loop of fine-tuning. The user interface transforms from a control panel into a dashboard for monitoring the agent's progress and reviewing its final, comprehensive report. This agentic framework dramatically accelerates the development of highly accurate, custom AI models, making production-grade GenAI more accessible and efficient.


python/agents/sft-runner-starter-pack/src/agent.py
```diff
@@ -0,0 +1,146 @@
...
+    callback_context.state["INITIAL_TARGET_EXAMPLES"] = config.initial_target_examples
+    callback_context.state["BQ_DATASET_ID"] = config.DATASET
+    print(f"Initial state setup complete.")
+
+
+# Root Agent
+root_agent = LlmAgent(
+    name="FineTuningCoordinator",
+    model=config.flash_model,
+    description="A master coordinator for a machine learning fine-tuning pipeline.",
```
mstyer-google: Maybe just "A coordinator"

python/agents/sft-runner-starter-pack/src/config.py
```diff
@@ -0,0 +1,87 @@
...
+        self.project_id: Optional[str] = os.getenv("GOOGLE_CLOUD_PROJECT") or os.getenv(
+            "GCP_PROJECT_ID"
+        )
+        self.location: Optional[str] = os.getenv(
+            "GOOGLE_CLOUD_LOCATION", "us-central1"
+        ).lower()
+        self.use_vertex_ai: bool = os.getenv("GOOGLE_GENAI_USE_VERTEXAI", "0") == "1"
+
+        # Bigquery configurations
+        self.BQ_LOCATION: Optional[str] = os.getenv("BQ_LOCATION") or os.getenv(
```
mstyer-google: This could be a Python trick I'm not familiar with, but why do you repeat `os.getenv("BQ_LOCATION")` here?

python/agents/sft-runner-starter-pack/src/config.py
```diff
@@ -0,0 +1,87 @@
...
+            "BQ_LOCATION"
+        )
+        self.PROJECT: Optional[str] = os.getenv("PROJECT") or os.getenv("PROJECT")
```
mstyer-google: Same here

python/agents/sft-runner-starter-pack/src/config.py
```diff
@@ -0,0 +1,87 @@
...
+        self.DATASET: Optional[str] = os.getenv("DATASET") or os.getenv("DATASET")
```
mstyer-google: And here

python/agents/sft-runner-starter-pack/src/config.py
```diff
@@ -0,0 +1,87 @@
...
+            raise ValueError("GOOGLE_CLOUD_PROJECT environment variable is required")
+        return True
+
+    @property
+    def project_location(self) -> str:
+        """Get the project location in the format required by BigQuery and Dataform."""
+        return f"{self.project_id}.{self.location}"
+
+    @property
+    def vertex_project_location(self) -> str:
```
mstyer-google: Doesn't this return the same string as the function above?

python/agents/sft-runner-starter-pack/src/agent.py
```diff
@@ -0,0 +1,146 @@
...
+    callback_context.state["eval_dataset_path"] = (
+        config.eval_dataset
+    )  # "data/sample_eval_queries.csv"
+    callback_context.state["project_id"] = config.project_id
+    callback_context.state["gcp_location"] = config.location
+    callback_context.state["gcs_bucket_name"] = config.bucket_name
+    callback_context.state["base_model"] = config.flash_model
+    callback_context.state["target_examples"] = config.initial_target_examples
+    callback_context.state["bq_dataset_id"] = config.DATASET
+    callback_context.state["GCP_PROJECT_ID"] = config.project_id
```
mstyer-google: I thin it's probably clearer to make the case consistent across all context keys. Personally, I would suggest all lowercase, unless there's a reason to make them all uppercase.

python/agents/sft-runner-starter-pack/src/agent.py
```diff
@@ -0,0 +1,146 @@
...
+)
+
+
+# --- State Setup ---
+def setup_before_agent_call(callback_context: CallbackContext):
+    """Sets up the initial state for the agent from environment variables."""
+    print("--- Setting up initial state from .config file ---")
+
+    callback_context.state["seed_data_path"] = (
+        config.seed_queries
```
mstyer-google, line 114: These context keys have confusing names. `config.seed_queries` and `context.eval_dataset` sound like the value would be the queries or the dataset, when IIUC they are actualy path strings.

python/agents/sft-runner-starter-pack/src/agent.py
```diff
@@ -0,0 +1,146 @@
...
+
+# Logging Setup
+logging.basicConfig(
+    level=logging.INFO, format="%(asctime)s - %(name)s - %(levelname)s - %(message)s"
+)
+logger = logging.getLogger(__name__)
+
+
+# Agent Definitions
+def create_data_generator_agent():
```
mstyer-google, line 54: Why do you use generator functions instead of just creating the agent inline as you do for `full_pipeline_agent`?

python/agents/sft-runner-starter-pack/src/prompts.py
```diff
@@ -0,0 +1,78 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Prompts definitions"""
+
+# Orchestrator prompt
+ROOT_PROMPT = """
+You are a master coordinator for a machine learning fine-tuning pipeline.
```
mstyer-google: Just use "a coordinator"

python/agents/sft-runner-starter-pack/LICENSE
mstyer-google, line 1: I don't think you need to include a copy of the Apache license in the repo.

python/agents/sft-runner-starter-pack/src/sub_agents/data_generator/tool.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+# Utility function: Generate synthetic data examples
+def _generate_examples_batch(
+    model, schemas_str, num_examples, existing_questions, sample_questions_df
+) -> str:
+    """Generate X number of synthetic examples"""
+    # TODO: Improve: Make the use case generic
+    prompt = f"""
+    You are an expert system designed to generate synthetic training data for fine-tuning text-to-SQL models.
```
mstyer-google, line 74: Isn't this relying on Gemini to generate SQL correctly in the first place? How will this sample data help fine-tune the model to improve SQL generation? I would imagine you'd want a set of "hard" queries (i.e. queries Gemini gets wrong), along with human-corrected SQL.

Also, what verifies that these generated examples are actually correct?

python/agents/sft-runner-starter-pack/src/sub_agents/evaluator/tool.py
```diff
@@ -0,0 +1,234 @@
...
+    )
+
+    # Define the evaluation task
+    logger.info("Defining the evaluation task...")
+    eval_task = EvalTask(
+        dataset=eval_df,
+        metrics=[
+            PointwiseMetric(
+                metric="relevance",
+                metric_prompt_template="""You are a professional writing evaluator. Your job is to score writing responses according to pre-defined evaluation criteria. You will be assessing question answering relevance, which measures the ability to respond with relevant information when asked a question. You will assign the writing response a score from 5, 4, 3, 2, 1, following the INDIVIDUAL RATING RUBRIC and EVALUATION STEPS.
```
mstyer-google: I don't think this eval prompt is asking for the right evaluation. This looks like a generic writing evaluation prompt. Don't you want the model to evaluate whether the SQL is correct?

mstyer-google (changes requested):
I made a few comments in the files. Also, a few general comments:

* Make sure to run the black autoformatter with `--line-length 80`, so it formats the lines correctly.
* Check all your files, class definitions and key methods have docstrings.

python/agents/sft-runner-starter-pack/src/agent.py
suddhasatwabhaumik: updated

python/agents/sft-runner-starter-pack/src/config.py
suddhasatwabhaumik: Removed. Was from an old/incorrect implementation

python/agents/sft-runner-starter-pack/src/config.py
suddhasatwabhaumik: Removed. Was from an old/incorrect implementation

python/agents/sft-runner-starter-pack/src/config.py
suddhasatwabhaumik: Removed. Was from an old/incorrect implementation

python/agents/sft-runner-starter-pack/src/config.py
suddhasatwabhaumik: Removed. Was from an old/incorrect implementation

python/agents/sft-runner-starter-pack/src/agent.py
suddhasatwabhaumik: Updated

python/agents/sft-runner-starter-pack/src/agent.py
suddhasatwabhaumik: Data Generator, Fine tuner, and Evaluator are individual sub agents which the users can run independently as well if they want, provided their pre-requisites already exists.

For example, the user already contains a fine tuned model and wants to evaluate, in which case, they will only ask the Agent to run the Evaluation. 

This is why I kept these 3 are LlmAgent() objects, while the Full pipeline Agent is a SequentialAgent() object. 

Please advise.

python/agents/sft-runner-starter-pack/src/agent.py
suddhasatwabhaumik: Both of them are paths in this version.
In a future version we can update them to BQ Tables/Datasets.

To make things simple for the users of this application to setup and run this tool, I provided this information in the Readme file.

python/agents/sft-runner-starter-pack/src/prompts.py
suddhasatwabhaumik: Updated

python/agents/sft-runner-starter-pack/src/sub_agents/data_generator/tool.py
suddhasatwabhaumik: We give Gemini a set of Seed queries (to be provided by the user of this application) to generate a synthetic set of X queries similar to the seed queries, where X is the number of examples users may want to generate. X should ideally be in 1000s. 

We use this variety of similarly looking SQL queries to Fine tune the model, but necessarily they are not evaluated, as the Seed queries are passed in the prompt along with the Table schema definition from BigQuery. Along with the Temperature settings used during this Gemini call, we restrict hallucinations to the best. However, we do not have an evaluation pipeline as of now to verify the SQL queries generated by Gemini. 

Overall, these queries are not the Human validated or LLM judged queries, but simply an oversampled set of SQL queries, for cases where customers do not have 1000s of queries already to fine tune a model and compare its performance with OOTB NL2SQL performance of Gemini 3, for example. 

Please advise if we should add a quick evaluator to the Gemini generated queries, to this particular step.

python/agents/sft-runner-starter-pack/src/sub_agents/evaluator/tool.py
suddhasatwabhaumik: Corrected.

python/agents/sft-runner-starter-pack/src/agent.py
mstyer-google: What I meant is, why do you create these agents using a separate function instead of just declaring them as objects like `full_pipeline_agent`?

Instead of defining a function called `create_generator_agent()`, you could just do:
```
generator_agent = LlmAgent(
    name="data_generator_agent",
    ...
```
Then in `root_agent` you would have:
```
root_agent = LlmAgent(
    ...
    sub_agents = [
        full_pipeline_agent,
        generator_agent,
        ...
    ]
```
Using generator functions instead adds a layer of abstraction & complexity I don't think you need. Someone reading the code needs to look at the function and figure out why you did it that way, and whether it's doing anything other than just creating the object. If there's a reason for it, that's great, but I don't think I see one.

python/agents/sft-runner-starter-pack/src/config.py
mstyer-google, line 65: I learned recently that setting temp=0 as a way to achieve deterministic results is outdated advice. The current (strong) recommendation for Gemini is temp=1 -- I would recommend just hard-coding the value here and not making it configurable. 

Using a fixed seed (which you do below) is a better way to get determinism.

python/agents/sft-runner-starter-pack/src/sub_agents/data_generator/tool.py
mstyer-google: It seems like you should evaluate the generated queries -- at a minimum to make sure they are syntactically correct. Beyond that, though, I wonder what you want the fine-tuning here to achieve. Aren't you just using Gemini's output to fine-tune Gemini? that seems circular.

mstyer-google (changes requested):

python/agents/sft-runner-starter-pack/src/agent.py
suddhasatwabhaumik: done

python/agents/sft-runner-starter-pack/src/config.py
suddhasatwabhaumik: done

happyhuman:
Hi @suddhasatwabhaumik, this PR has been open for **110** days without recent activity. Some of the files it touches have been restructured on `main` in the interim. Are you still interested in continuing this? If we don't hear back within the next **14 days**, we'll close this PR for housekeeping — please feel free to reopen anytime if you'd like to pick it back up. Thanks for the contribution!

_(auto-managed by repo housekeeping — nudge id: `cloudcode-2wk-nudge-2026-07-28`)_

happyhuman:
Hi @suddhasatwabhaumik, closing this for housekeeping — we checked in 2 weeks ago and haven't heard back. This isn't a judgement on the change itself: please feel free to reopen anytime if you'd like to pick it back up, and thanks again for the contribution!

(auto-managed by repo housekeeping — nudge id: cloudcode-2wk-nudge-2026-07-28, close id: cloudcode-2wk-close-2026-08-12, session: ses_0090a2a81ffe94zOTLxffBdenD)

PR #815, opened 2026-01-06
product catalog ad generation agent update

tkhouw:
Updates the product catalog ad generation agent
- Uses newest available models for significant gains in output quality
- Allows users to provide resources to the agent (such as product images and asset sheets) directly through the ADK web UI without relying on cloud storage
- Prompt optimization
- Some code refactoring

### DIFF
--- a/python/agents/product-catalog-ad-generation/scripts/generate_sample_data.py
+++ b/python/agents/product-catalog-ad-generation/scripts/generate_sample_data.py
@@ -28,7 +28,7 @@ from pydantic import BaseModel, Field
 load_dotenv()
 
 PROJECT = os.getenv("GOOGLE_CLOUD_PROJECT")
-LOCATION = os.getenv("REGION", "us-central1")
+LOCATION = os.getenv("GCP_LOCATION", "us-central1")
 COMPANY_NAME = os.getenv("COMPANY_NAME", "ACME Corp")
 
 # Output directories


python/agents/product-catalog-ad-generation/agent_engine_deploy.py
```diff
@@ -218,12 +218,15 @@ def create() -> None:
         display_name=root_agent.name,
```
happyhuman: Can we modify create(), so it reads the required packages from a requirements.txt?

python/agents/product-catalog-ad-generation/content_gen_agent/func_tools/generate_audio.py
```diff
@@ -113,33 +62,109 @@ async def generate_audio(
         logging.error("GOOGLE_CLOUD_PROJECT environment variable not set.")
```
happyhuman, line 62: I see GOOGLE_CLOUD_PROJECT in some places and GOOGLE_CLOUD_PROJECT_ID in other files. Is this intentional, or maybe a bug?

Similarly, I noticed both GOOGLE_CLOUD_LOCATION_REGION and GCP_LOCATION variables in some other files, but they may or may not be meant to be the same thing.

python/agents/product-catalog-ad-generation/agent_engine_deploy.py
tkhouw: We realized that this deploy script has become obsolete, so we have decided to remove it

python/agents/product-catalog-ad-generation/content_gen_agent/func_tools/generate_audio.py
tkhouw: Good catch!  As noted above, the agent_engine_deploy.py script was actually obsolete, so I have just removed it.  I don't believe the environment variable names are currently inconsistent in any other files.

PR #834, opened 2026-01-14
<feat>(GH action): ruff githook

tommywagz:
Leverages a github action through a yml workflow. Doesn't actively fix ruff check errors on behalf of developers - only highlights them to be fixed manually. The simple fixes can be automated if allowed, but the isort formatting is automatic.


python/agents/RAG/pyproject.toml
happyhuman, line 1: Let's not fix a sample in this PR.

.github/workflows/ruff-checks.yml
happyhuman, line 1: Let's rename checks.yml to python-checks.yml or ruff-checks.yml (in case other languages also add their own checkers later).

happyhuman (approved):

PR #840, opened 2026-01-14
<fix>(ruff): auto insurance agent ruff changes

tommywagz:
auto insurance agent ruff .toml updates


python/agents/auto-insurance-agent/auto_insurance_agent/__init__.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from . import agent  # noqa: F401
+from . import agent
```
happyhuman, line 15: Let's keep it as is (I explained why in a comment for https://github.com/google/adk-samples/pull/841)

PR #841, opened 2026-01-14
<fix>(ruff): ruff changes for blog writer

tommywagz:


python/agents/blog-writer/blogger_agent/sub_agents/__init__.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .blog_editor import blog_editor  # noqa: F401
-from .blog_planner import robust_blog_planner  # noqa: F401
-from .blog_writer import robust_blog_writer  # noqa: F401
-from .social_media_writer import social_media_writer  # noqa: F401
+from .blog_editor import blog_editor
```
happyhuman, line 15: I see F401 is globally turned off in the root/pyproject.toml
I don't think we want it completely turned off. There are often unused imports that we want to remove. The only ones I have come to notice are the ones in __init__.py files that sometimes black/ruff might find unused, but they are separately used in other files. So, let's undo these 4 lines, and also modify the following in root/pyproject.toml:
```
# OLD:
[tool.ruff.lint.per-file-ignores]
# Global rule: ADK agents use __init__.py to expose their logic
"__init__.py" = ["F401"]
```
to this:
```
# NEW (from root/.flake8) :
[tool.ruff.lint.per-file-ignores]
# Global rule: ADK agents use __init__.py to expose their logic
"__init__.py" = ["E203", "W503", "E501", "E704", "E701"]
```

PR #842, opened 2026-01-14
<chore>(ruff): currency agent ruff changes

tommywagz:


python/agents/currency-agent/currency_agent/__init__.py
```diff
@@ -1 +1 @@
-from . import agent  # noqa: F401
+from . import agent
```
happyhuman, line 1: Let's keep it as is (I explained why in a comment for https://github.com/google/adk-samples/pull/841)

PR #1019, opened 2026-02-04
feat: Add Navallist Lite agent sample

tpryan:
Initial import of Navallist Lite: a single-binary Go application implementing an embedded ADK agent, HTTP server, realtime service, and SQL-backed store. Adds cmd/server entrypoint, internal packages (agent, config, data, realtime, server), disk artifact storage, migrations, frontend assets (vanilla JS/CSS/HTML), OpenAPI and checklist schema docs, Makefile and .env.example for local dev, plus go.mod/go.sum and tests. Development notes: use `make db-start` to launch a local DB and `make run` to run the server.

### DIFF
--- a/python/agents/RAG/uv.lock
+++ b/python/agents/RAG/uv.lock
@@ -8,7 +8,7 @@ resolution-markers = [
 
 [[package]]
 name = "agent-starter-pack"
-version = "0.32.0"
+version = "0.35.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "backoff" },
@@ -19,9 +19,9 @@ dependencies = [
     { name = "requests" },
     { name = "rich" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/6f/ad/4ddd9d52bc9c64621c660227c2b686712000151e6d95edb57bc0fc243bb6/agent_starter_pack-0.32.0.tar.gz", hash = "sha256:623a3411b35464b16c1979fcc164723163e2c60fb3ec07487030d1260518917d", size = 11868476, upload-time = "2026-01-30T19:11:18.091Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/fa/df/991a1d97ebf81bf367e045cb5c47465ba90813d64ccf0dd2044085531528/agent_starter_pack-0.35.1.tar.gz", hash = "sha256:9940e49a0f8cfe6fe2a08ae85db3369cace31a6b8a3f96de75fdd1cbd05286bb", size = 11981504, upload-time = "2026-02-10T10:32:38.687Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/12/37/6467c853c9e0515ff49a0017affcb6a2ab5f5954ea086b3d6c66641c20d5/agent_starter_pack-0.32.0-py3-none-any.whl", hash = "sha256:22e480a22436c335b09ec517d48f00d6a9a9445ed1ea05a154e99712bd37c193", size = 3191784, upload-time = "2026-01-30T19:11:20.688Z" },
+    { url = "https://files.pythonhosted.org/packages/da/be/a7298e93adfe265611eccc36e3b4e896556275ce5996555994c3b96d5fc5/agent_starter_pack-0.35.1-py3-none-any.whl", hash = "sha256:b4539f050c38b94b0a6aada261a4dcb6a5575c67cba3b3bd39f8a02da1240c30", size = 3315119, upload-time = "2026-02-10T10:32:41.119Z" },
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


go/agents/navallist/web/js/app.js
```diff
@@ -0,0 +1,961 @@
...
+                );
+                
+                console.log("Agent Replied:", events);
+
+                // Check for "I cannot assign" messages to surface to user
+                if (Array.isArray(events)) {
+                    events.forEach(ev => {
+                        if (ev.content && ev.content.parts) {
+                            ev.content.parts.forEach(p => {
+                                if (p.text && p.text.includes("I cannot assign")) {
```
ivanmkc: Should probably have some sort of isWarning: boolean, or enumeration instead of hardcoding a magic string.

go/agents/navallist/web/js/app.js
```diff
@@ -0,0 +1,961 @@
...
+                                }
+                            });
+                        }
+                    });
+                }
+
+                ui.showToast("Got it!", "success");
+            } catch (e) {
+                console.error(e);
+                if (e.message.includes("overloaded")) {
```
ivanmkc: Ditto. Consider having the status have a dedicated field in the response instead of inferring from the message.

go/agents/navallist/web/js/app.js
```diff
@@ -0,0 +1,961 @@
...
+        });
+        
+        const audioBlob = await media.stopRecording();
+        if (audioBlob) {
+            console.log("Sending Audio...");
+            try {
+                // Send "Process this" prompt with audio
+                // Inject identity so the agent knows who is speaking
+                const prompt = `This is ${state.displayName} speaking. Process this voice command.`;
+                const events = await api.sendInteraction(
```
ivanmkc, line 904: nit: For a production use case, this can be fragile since the network response may take a while to respond (seconds or tens of seconds) due to the LLM call. The device/carrier (especially mobile over cellular) may reset the connection after mere seconds, especially if a user switches app mid-flight. 

For sample purposes, what you have is probably fine, especially if you explain in comments/documentation the potential issues of this design.

For comprehensiveness, here's a proposal for moving to a more robust UI, which uses events and websockets to remove reliance on the open HTTP connection:

# Asynchronous Agent Execution Plan

The goal is to fix the blocking UX by refactoring the `/api/agent/run` endpoint and the vanilla JS frontend to use a non-blocking, event-driven pattern. The interactions will move to the background, and agent text responses will flow back via WebSockets.

## The Problem

In the current architecture, when the frontend sends an interaction to the agent (e.g., a voice command or text), it makes a blocking HTTP `POST` to `/api/agent/run`.

1. **Poor UX & Latency:** The HTTP connection stays open while the LLM thinks, formats, and executes tools. This can take 3–10 seconds, during which the user is staring at a loading state on the UI.
2. **Network Fragility:** Mobile browsers are aggressive about terminating long-lived pending HTTP requests if the network fluctuates or if the app is backgrounded.
3. **Redundant Flow:** The frontend actually updates its visual state earlier than the HTTP response arrives, because the backend tools update the database—triggering a WebSocket data push. The browser parses the final HTTP response strictly to look for edge case warning messages (e.g. "I cannot assign this").

## Proposed Changes

### 1. Backend: HTTP Handler & Background Worker
Modify the Go server to instantly return a 202 Accepted, moving the LLM execution logic into a Goroutine.

#### [MODIFY] `internal/server/handlers/checklist.go`
- Update the `RunInteraction` method.
- Currently, it reads the input, calls `AgentClient.Run`, and then JSON encodes the response back to the writer.
- **Change:** Parse the input, launch a `go func() { ... }` passing a detached context (like `context.Background()` or a custom timeout context so it survives the HTTP request closing).
- **Change:** Immediately call `w.WriteHeader(http.StatusAccepted)` on the main thread and return.

### 2. Backend: WebSocket Broadcasting for Agent Responses
The agent's text response (e.g., "I cannot assign this...") needs a path back to the browser.

#### [MODIFY] `internal/server/handlers/checklist.go` (or a worker)
- Inside the Goroutine, after `AgentClient.Run` completes, inspect the returned events.
- If there are user-facing messages (e.g., the `"I cannot assign..."` error), construct a simple JSON payload.
- Use the existing `realtime.Service` (or `Hub`) to publish an `agent_notification` event to the specific `tripId` channel.

### 3. Frontend: Fire-and-Forget
Stop keeping the HTTP connection alive and remove the parsing logic from the frontend request.

#### [MODIFY] `web/js/api.js`
- Update `sendInteraction()` to simply check `res.ok` (to confirm the 202 Accepted) and return. It no longer needs to wait for or parse the heavy JSON response payload.

#### [MODIFY] `web/js/app.js`
- Locate the `api.sendInteraction` call inside the microphone/voice handler (`handleCameraClick` / Voice logic around line ~900).
- Remove the block of code that iterates over `events.forEach(...)` looking for `"I cannot assign"`, as the HTTP call will no longer return the events list.

### 4. Frontend: Listen for Native WebSocket Messages
The UI needs to listen to the WebSocket for the agent's text responses.

#### [MODIFY] `web/js/app.js`
- Find the realtime subscription (`subscribeTrip` callback block) which currently listens for standard trip data changes.
- Add a new `else if (type === 'agent_notification')` handler.
- When an `agent_notification` is received, invoke `ui.showToast(data.message, data.level)`.

## Verification Plan
### Manual Verification
1. Run the local database (`make db-start`) and Go app (`make run`).
2. Open the browser and test checking an item manually. It should sync quickly without issues.
3. Test a Voice interaction using the microphone button ("Assign the engine oil to a fake user").
4. Verify that the UI instantly frees up, and 3-5 seconds later, a UI Toast surfaces stating the assignment failed, delivered exclusively via the WebSocket.

go/agents/navallist/internal/agent/agent.go
```diff
@@ -0,0 +1,102 @@
...
+			Name:        "update_checklist_item",
+			Description: "Updates the status and location of a specific item on the boat's checklist. You can only assign items to names exactly as they appear in 'get_crew_list'. Assigning to anyone else is forbidden.",
+		},
+		handler.UpdateItem,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	bulkTool, err := functiontool.New(
```
ivanmkc: I have a similar setup in my MCP server, but just use a single bulk_update tool, which should be applicable for the single-item use case as well. I would recommend keeping this Tool and removing `updateTool`

go/agents/navallist/internal/agent/agent.go
```diff
@@ -0,0 +1,102 @@
...
+			Name:        "bulk_update_items",
+			Description: "Updates multiple checklist items at once. Use this when the user says 'check all my items' or gives a list of updates. This is much more efficient than calling update_checklist_item multiple times.",
+		},
+		handler.BulkUpdateItems,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	crewTool, err := functiontool.New(
```
ivanmkc, line 45: Instead of having the agent decide if need to call this tool, taking up time and tokens, consider pre-emptively injecting this context into the agent instructions if it's small enough.

Same for "get_checklist_status".

go/agents/navallist/internal/agent/client.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	if err := json.Unmarshal(b, &req); err != nil {
+		return nil, fmt.Errorf("invalid payload structure: %w", err)
+	}
+
+	if req.AppName == "" || req.UserID == "" || req.SessionID == "" {
+		return nil, fmt.Errorf("app_name, user_id, session_id are required, got app_name: %q, user_id: %q, session_id: %q", req.AppName, req.UserID, req.SessionID)
+	}
+
+	if req.AppName != c.Agent.Name() {
```
ivanmkc: Something missing here?

go/agents/navallist/internal/agent/agent.go
tpryan: Fixing.

go/agents/navallist/internal/agent/agent.go
tpryan: It's pretty short lived and people are being added , so querying to get the most up to date list is I think the better path forward.

go/agents/navallist/internal/agent/client.go
tpryan: ¯\(°_o)/¯ Removing.

go/agents/navallist/web/js/app.js
tpryan: I'm going to chose to just give a warning as frontend of this is secondary to the sample.  But man this is a good catch, and I'll add it to my real version of this app.

go/agents/navallist/web/js/app.js
tpryan: fixed.

go/agents/navallist/web/js/app.js
tpryan: fixed.

PR #1089, opened 2026-02-15
Add Agent Starter Pack deployment option for data-engineering agent sample

maeve2024:
## Summary

Add Agent Starter Pack as a supported deployment option for the data-engineering agent.

This includes migrating to `uv`, updating dependencies, and adding supporting configuration for development and packaging.

## Changes

- Add Agent Starter Pack as an alternative project/deployment workflow
- Migrate from Poetry to `uv`
- Switch build backend to `uv_build`
- Update `google-adk` to `>=1.0.0`
- Regenerate `uv.lock`
- Update Python version constraint to `>=3.10,<3.13`
- Add dev dependencies (`agent-starter-pack`, `nest-asyncio`)
- Add lint/type tooling configuration (`ruff`, `mypy`, `codespell`)
- Update README with `uv` and Agent Starter Pack instructions
- Minor config cleanup and formatting updates

## Testing

image
image

- Verified `uv sync` installs successfully
- Ran tests using `uv run pytest`
- Tested locally with ADK Web UI
- Confirmed starter pack workflow initializes correctly

### DIFF
--- a/python/agents/data-engineering/data_engineering_agent/__init__.py
+++ b/python/agents/data-engineering/data_engineering_agent/__init__.py
@@ -11,14 +11,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 import os
 
 import google.auth
+
+from .agent import root_agent
+
+__all__ = ["root_agent"]
 
 _, project_id = google.auth.default()
 os.environ.setdefault("GOOGLE_CLOUD_PROJECT", project_id)
 os.environ.setdefault("GOOGLE_CLOUD_LOCATION", "global")
 os.environ.setdefault("GOOGLE_GENAI_USE_VERTEXAI", "True")
-
-from . import agent
--- a/python/agents/data-engineering/data_engineering_agent/config.py
+++ b/python/agents/data-engineering/data_engineering_agent/config.py
@@ -26,20 +26,16 @@ if env_path.exists():
 class Config:
     def __init__(self):
         # Google Cloud Configuration
-        self.project_id: str | None = os.getenv(
-            "GOOGLE_CLOUD_PROJECT"
-        ) or os.getenv("GCP_PROJECT_ID")
+        self.project_id: str | None = os.getenv("GOOGLE_CLOUD_PROJECT") or os.getenv(
+            "GCP_PROJECT_ID"
+        )
         self.location: str | None = os.getenv(
             "GOOGLE_CLOUD_LOCATION", "us-central1"
         ).lower()
-        self.use_vertex_ai: bool = (
-            os.getenv("GOOGLE_GENAI_USE_VERTEXAI", "0") == "1"
-        )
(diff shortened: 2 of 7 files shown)


python/agents/data-engineering/pyproject.toml
```diff
@@ -6,10 +6,8 @@ authors = [
     { name = "Samet Karadag", email = "sametkaradag@google.com"},
     { name = "Saurabh Maurya", email = "saurabhmaurya@google.com"}
 ]
-license = {text = "Apache License 2.0"}
+license = "Apache-2.0"
```
zeroasterisk, line 9: Is this hand coded? Where did this change come from?

python/agents/data-engineering/pyproject.toml
```diff
@@ -20,31 +18,88 @@ dependencies = [
...
+    "mypy>=1.15.0",
+    "codespell>=2.2.0",
+    "types-pyyaml>=6.0.12.20240917",
+    "types-requests>=2.32.0.20240914",
+]
+
 [tool.ruff]
-extend = "../../../pyproject.toml"
+line-length = 88
+target-version = "py310"
```
zeroasterisk, line 48: Is this too forceful or is it fine?

python/agents/data-engineering/README.md
```diff
@@ -31,16 +31,15 @@ A comprehensive data engineering agent.
 ## Setup and Installation
 
 1.  **Prerequisites**
+    *   Python 3.10+
```
zeroasterisk, line 34: Should we recommend the make commands instead of the direct UV commands?

python/agents/data-engineering/README.md
```diff
@@ -184,9 +183,37 @@ Response:
 To delete the deployed agent, you may run the following command:
 
 ```bash
-python3 deployment/deploy.py --delete --resource_id=${AGENT_ENGINE_ID}
+uv run deployment/deploy.py --delete --resource_id=${AGENT_ENGINE_ID}
+```
+
+### Alternative: Using Agent Starter Pack
+
+You can also use the [Agent Starter Pack](https://goo.gle/agent-starter-pack) to create a production-ready version of this agent with additional deployment options:
```
zeroasterisk, line 191: I think we should start with the agent starter pack for all of these. It is the easiest way to get going and there's no reason not to make it easy for our users. We should provide alternative commands if they choose not to use the agent starter pack, but let's get started with the one-liner

python/agents/data-engineering/pyproject.toml
maeve2024: Yes it is hard coded. Many build and publishing tools (uv) require the exact SPDX license identifier to recognize and validate the license. It was in Elia's PR as well: https://github.com/google/adk-samples/pull/351/changes.

python/agents/data-engineering/pyproject.toml
maeve2024: You're right. I'll change this back since it's a pattern with the other data agents to point back to the master template.

python/agents/data-engineering/README.md
maeve2024: We can. Elia's PR switched from poetry to uv -- so i based it off that.

PR #1098, opened 2026-02-17
fix: harden file handling for brand_aligner_agent

mohabfekry:
Move GCS URI construction to tools rather than agent instructions, further reducing cognitive load and ensuring reliable GCS file logic.


mohabfekry:
Will redo from a fresh fork to avoid already merged commits

PR #1112, opened 2026-02-18
docs(rag): add pip install fallback to README

ApurvaPatil2401:
While testing the RAG sample, I encountered an error because the README only provided uv sync instructions, and there was no requirements.txt. I have verified that pip install . works correctly as a fallback for users not using uv. Updated the README to include both options.

### DIFF
--- a/python/agents/RAG/README.md
+++ b/python/agents/RAG/README.md
@@ -56,10 +56,15 @@ This diagram outlines the agent's workflow, designed to provide informed and con
 
 2.  **Install Dependencies:**
 
+    If you have `uv` installed (recommended):
     ```bash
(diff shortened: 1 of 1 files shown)


happyhuman:

ApurvaPatil2401:
Thanks @happyhuman! I checked the updated recipe structure under core/ and saw that the setup process now uses make commands (make install).

happyhuman:
That is not necessarily a requirement. I suggest that you look at the [docs](https://github.com/google/adk-samples/tree/main/docs) to learn more about the new quality standards and requirements for the samples/recipes, should you choose to submit a recipe.

PR #1157, opened 2026-02-24
docs: rename ADK Bidi-streaming to ADK Gemini Live API Toolkit

kazunori279:
- Update project name across all files
  - Fix duplicate thinking text in chat bubbles
  - Update screenshot with new UI
  - Modernize pyproject.toml (SPDX license, optional-dependencies)


python/agents/bidi-demo/app/static/js/app.js
```diff
@@ -668,6 +671,17 @@ function connectWebsocket() {
...
+          // Skip thinking/reasoning text from chat bubbles (shown in event console)
+          if (part.thought) {
+            hasThoughtInTurn = true;
+            continue;
+          }
+
+          // Skip final aggregated content that replays thinking text already seen
+          if (!adkEvent.partial && hasThoughtInTurn) {
+            continue;
+          }
```
koverholt: Could this filter inadvertently drop actual response text? `partial` isn't always set on ADK events, so `!adkEvent.partial` can be true for normal response events too.

Would the previous `part.thought` check alone be sufficient here?

koverholt (approved):
LGTM, thanks for the updates!

I left one comment to check the second thought filter, but feel free to ignore if you already tested this!

python/agents/bidi-demo/app/static/js/app.js
kazunori279: Thanks a lot for finding this! `part.thought` isn't enough since the ADK sends a final aggregated content event that replays the same thinking text without the thought flag set (at least with the native-audio model), which caused a duplicate. So replaced `hasThoughtInTurn` with `hasOutputTranscriptionInTurn` to avoid dropping the resp text.

PR #1284, opened 2026-03-17
feat(youtube-analyst): add Agent Starter Pack support

maeve2024:
REVISION:
- Added # noqa: E402 on the from .agent import root_agent line so ruff allows the non-top import. 

CHANGES:
- Add [tool.agent-starter-pack] and agent-starter-pack dev dep so the agent can be used as an ASP template.
- Add flexible auth in init.py (google.auth.default + env setdefault) so it works with ADC, service accounts, and Cloud Run/GKE without hardcoding project/location.
- Make ASP the main path in README and recommend uv one-liner first so users get a production-ready setup by default; clone-and-run is the fallback.
- Remove load_artifacts from the root agent so ASP-generated projects pass integration tests when the Runner has no artifact service (avoids "Artifact service is not initialized").
- Add pytest-asyncio and asyncio_mode = "auto" so generated projects’ async tests run without extra config.
Regenerate uv.lock for new dev deps.

TESTS:
image
image

### DIFF
--- a/python/agents/youtube-analyst/README.md
+++ b/python/agents/youtube-analyst/README.md
@@ -33,95 +33,33 @@ graph TD
 
 ## Agent Details
 
-| Feature | Description |
-| --- | --- |
-| **Interaction Type:** | Conversational / Analytics |
-| **Complexity:**  | Intermediate |
-| **Agent Type:**  | Multi-Agent (Root + Visualization Sub-Agent) |
-| **Components:**  | YouTube Data API, ADK Tools, Interactive Plotly Charts |
-| **Vertical:**  | Marketing / Media Analytics |
+| Feature               | Description                                            |
+| --------------------- | ------------------------------------------------------ |
+| **Interaction Type:** | Conversational / Analytics                             |
+| **Complexity:**       | Intermediate                                           |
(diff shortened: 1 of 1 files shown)


python/agents/youtube-analyst/youtube_analyst/agent.py
```diff
@@ -2,7 +2,6 @@
 
 from google import genai
 from google.adk.agents import Agent
-from google.adk.tools import load_artifacts
```
zeroasterisk, line 5: Please explain why this is removed?

zeroasterisk: Answered in conversation 

> Remove load_artifacts from the root agent so ASP-generated projects pass integration tests when the Runner has no artifact service (avoids "Artifact service is not initialized").

Should we instead add an artifact service.

@tpryan do you have an opinion on this one?

tpryan: To make sure I understand first:
youtube sample use the artifact service. 
our test runner cannot run it because the artifact service is not enabled at the project level. 
Can you point me to what would have to be done to enable that?

PR #1286, opened 2026-03-18
feat: add agent starter pack.

maeve2024:
Updated https://github.com/google/adk-samples/pull/1174 with suggested revisions. 

Integration tests pass in an ASP-generated project; agent manually tested in ADK web UI (make playground).

PR #1337, opened 2026-03-25
Feat/youtube analyst agent starter pack

maeve2024:
- Add .env template and align YouTube Analyst README with sample standard

- Added `.env.example` so users can copy a working template instead of creating `.env` from scratch, matching the pattern used by other agent samples.

- Updated the README structure/content to align with PR https://github.com/google/adk-samples/pull/1286.

-Also, load_artifacts was dropped from the main YouTube agent in the Agent Starter Pack update so it isn’t duplicated there, while the visualization sub-agent still keeps it for workflows that need uploaded files.


zeroasterisk (changes requested):
Almost there:

- Bump `agent-starter-pack` from `>=0.14.1` to `>=0.32.0`

Rest of the PR looks solid.

PR #1340, opened 2026-03-26
fix: remove __import__ from sandbox builtins in policy-as-code agent

### LINKED ISSUE #1339: __import__ available in policy-as-code sandbox builtins
The `safe_globals` builtins dict in `simulation.py` includes `__import__`, making it accessible to exec'd code via `__builtins__['__import__']`. The AST validator blocks direct `__import__()` calls but not subscript access.

vnykmshr:
Fixes #1339

Removes `__import__` from the sandbox builtins dict. The AST validator catches direct calls but not subscript-style access via the builtins dict.

json/re/datetime are already injected into the sandbox namespace so import statements aren't needed in generated code.

Test added.


happyhuman:

PR #1409, opened 2026-04-03
Configure separate region for RAG engine

pemujo:
This PR enhances the RAG Agent setup flow by improving region flexibility for RAG engine and streamlining the IAM permission model.

Changes:

- GCP Region Decoupling: Changed the default region for the RAG Engine from us-central1 to us-west1 to avoid common allowlist restrictions.

  - Introduced an interactive prompt in the Makefile to allow users to specify their preferred region.

  - Decoupled the RAG Engine location from the Agent Engine location to allow for cross-region architectures.

- Simplified IAM Permissions: Removed the automated creation of the custom ragCorpusQueryRole.

  - Migrated to using the pre-defined roles/aiplatform.viewer role, which contains the necessary aiplatform.ragCorpora.query permissions out of the box.

- Improved Setup UX: Updated the setup_corpus target with more descriptive headers and a default-to-Yes [Y/n] logic for sample data deployment.


happyhuman:

PR #1441, opened 2026-04-07
[Small business Loan Agent] Facilitate 1 click deployment

norarfilali:
support deployment in Gemini Enterprise, remove env variables for 1 click deployment with agent garden, add fallback to upload file from GCS


tommywagz (changes requested):
Make sure that you inherit the .toml ruff rules from the repo's root .toml file. use these lines to do it:

```
[tool.ruff]
# extends the ruff rules established in the root .toml
extend = "../../../pyproject.toml"
```

Ping me once you add this bit

tommywagz (changes requested):
The agent looks good, I like the structure you have here.

Only one more thing I need, and that's 2 different authors on the .toml file. Even if no one else contributed to it. 

The rest of it is perfect, as soon as you add another author I'll merge

norarfilali:
Done @tommywagz - I added an additional author. Please review

tommywagz (approved):
LGTM! Thanks for the changes!!

PR #1485, opened 2026-04-10
[For on-brand-genmedia] Updated pyproject.toml with agent_starter_pack

anushasabbineni1:
Updated pyproject.toml with agent_starter_pack


happyhuman (commented):
We have recently restructured the repo and introduced a number of standardizations described in our [docs](https://github.com/google/adk-samples/tree/main/docs). Python samples (which are now called recipes) should be moved to contrib/python. Once you create a PR, please keep an eye on the workflows and run them if you need to and make sure they all pass without any errors. We will be able to review your PR after that.

PR #1486, opened 2026-04-10
Feat/airflow version upgrade agent

paraschavda2411:
Updates :
1. Added single click deploy functionality using Makefile
2. Updated couple of tools and prompts to support API way of deployment with enhancements.


tommywagz (approved):
LGTM! The .env gcloud config configuration was clever

PR #1503, opened 2026-04-11
chore: updated README & set default location for deployment.

cc4i:
- Added instruction for Agent Engine deployment in README
- The default location is us-central1 for deployment to Agent Engine
- Enable telemetry collection by default


tommywagz (approved):
There are a couple ruff checks that need fixing but they aren't too consequential, I approved if you want to merge. 

If you want to fix them with 
`uv run ruff format .`
and 
`uv run ruff check --fix`

If you do choose to make those changes I'll merge those

cc4i:
I run "uv run ruff check ." before commit, not sure if CI has stricter settings, fixed, thanks.

PR #1551, opened 2026-04-14
Feat/asp product catalog ad generation

MoisesRBitton:
This PR completes the Agent Starter Pack migration for the product-catalog-ad-generation agent and includes the runtime fixes needed so the migrated agent works reliably in ADK Web and local runs.

What this PR changes

Aligns the package entrypoint and exports with ASP expectations.
Updates project dependencies and config to the ASP-compatible stack.
Regenerates the lockfile with the resolved dependency graph.
Adds a smoke test for ADK app wiring using App plus InMemoryRunner.
Updates README to an ASP-first and uv-first setup flow.
Standardizes environment handling and model configuration through env variables.
Fixes function schema compatibility issues for automatic function calling.
Improves type-safety and SDK-compatibility in media and generation tool paths.
Hardens storyline JSON parsing with a repair fallback when model output is malformed.
Keeps runtime behavior resilient when optional external data sources are unavailable.
Why extra runtime fixes are included
During migration, startup and execution blockers appeared that prevented normal ADK Web flows. These fixes are included in the same PR so the migration is usable end-to-end, not only syntactically migrated.

image
image
@zeroasterisk


zeroasterisk (changes requested):
there's just way too many changes in this PR.  Sorry.  Can we reduce to just ASP. Then make a second PR with all the rest of the changes you might want to make.

happyhuman:

PR #1571, opened 2026-04-15
fix(invoice-processing): add build-system and agent-starter-pack conf…

tanyagoogle:
Without a [build-system] section, uv sync never installs invoice_processing as an importable package, causing ModuleNotFoundError when deployed via agent-starter-pack. Also adds explicit agent-starter-pack configuration, agent-catalog keywords, and ruff per-file-ignores to match the pattern established by nurse-handover.


tommywagz (approved):
LGTM

PR #1572, opened 2026-04-15
chore: Fix linter violations across the repo

zeroasterisk:
This PR resolves outstanding Ruff violations (B008, B904, B905, E402, RUF005, etc.) to achieve a zero-error linter state.

### DIFF
--- a/core/python/genmedia-for-commerce/tests/conftest.py
+++ b/core/python/genmedia-for-commerce/tests/conftest.py
@@ -19,6 +19,7 @@ import shutil
 from pathlib import Path
 
 import pytest
+from PIL import Image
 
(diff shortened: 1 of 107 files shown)


happyhuman:
Thanks for the PR. Please note that we just introduced a lot of quality standards and the new linting standards are defined in [pyproject.toml](https://github.com/google/adk-samples/blob/main/pyproject.toml) and we are trying to universally apply those rules to all the samples/recipes. If you have a strong reason that the linting rules in this file needs an update, please share it with me, so we can decide.
There are also some github workflows that have failed on this PR by the way, which need your attention. Please let me know if you have any questions.

PR #1604, opened 2026-04-15
fix(invoice-processing): regenerate uv.lock to enable agent-starter-p…

tanyagoogle:
The uv.lock was missing the agent-starter-pack dependency, which prevented the starter pack's version-lock mechanism from triggering during deployment. This caused `make register-gemini-enterprise` to fail because v0.15.4 lacks that Makefile target. Regenerating the lock file resolves agent-starter-pack to v0.41.0, matching the pattern used by other agents.

Also adds [[tool.uv.index]] for consistent PyPI resolution.


tommywagz (approved):
LGTM!

PR #1617, opened 2026-04-16
Fix/invoice processing agent engine project

tanyagoogle:
## Summary
      - Learning path fails with "PROJECT_ID not set" on Agent Engine because `config.py` and `alf_engine.py` capture
      `os.getenv("PROJECT_ID")` at module import time — before Agent Engine injects env vars
      - Agent Engine also never sets `PROJECT_ID`; it only provides the project via `GOOGLE_CLOUD_PROJECT` or
      `google.auth.default()`
      - Replaced module-level `LLM_*` variables with lazy getter functions that defer reads to call time, check 4 env var names,
      and fall back to `google.auth.default()` — matching the inference path pattern


tommywagz (approved):
LGTM!

PR #1661, opened 2026-04-17
uv upgrade and central1-pivot

tommywagz:
Removed broken 2.0 samples and upgraded uv across the board and changed all central1 to east1


tommywagz:
Didn't fetch

PR #1733, opened 2026-04-27
Use uv to run adk

tylere:
The current README instructions show how to build a UV environment that includes ADK using `uv sync`, but they don't describe how to use that environment to run ADK. These changes show how to run ADK using the uv-created environment.

(An alternative approach would be to activate the virtual environment before running the ADK command.)


happyhuman:

PR #1766, opened 2026-04-30
Claim adjudication updates

sandipgupta-oss:
Performed following changes
1) ADK to 1.31.1
2) Region to us-east1
3) Some doc changes


happyhuman:
Please take a look at the tests/checkers that are failing (you can skip the ruff formatter one).

PR #1767, opened 2026-04-30
Updated region and version for adk

paraschavda2411:
Changed regions to us-east1 and updated google adk package version


happyhuman:

PR #1790, opened 2026-05-01
fix: remove erroneously added files

haranrk:


koverholt (approved):
Thanks for catching this, cleanup LGTM!

PR #1917, opened 2026-05-12
feat(contrib/python): add crypto-payroll-agent recipe

plagtech:
## Summary
 
Adds `crypto-payroll-agent`, a sample that batch-pays multiple
recipients in ETH or ERC-20 tokens on Base in a single transaction. The
agent uses the Spraay batch payment tools from
[`google-adk-community`](https://github.com/google/adk-python-community/tree/main/src/google/adk_community/tools/spraay).
 
## Pattern demonstrated
 
- Tool selection across four related tools: the agent picks among
  `spraay_batch_eth`, `spraay_batch_token`, and their `_variable`
  siblings based on whether the user wants equal or variable amounts
  and ETH or ERC-20.
- Tool composition: a local `split_pool_proportionally` helper feeds
  the variable batch tools when the user supplies a total and weights
  rather than per-recipient amounts.
- A local `lookup_token_info` helper resolves common Base tokens
  (USDC, WETH, cbBTC, DAI, etc.) so users don't need to know decimals.
- Human-in-the-loop confirmation with a configurable USD ceiling.
 
## What's included
 
- `crypto_payroll_agent/` — single `LlmAgent` with the four Spraay
  tools plus two local helpers
- `eval/` — five-case evalset covering each batch-tool selection path
  and a safety refusal
- `tests/` — 14 offline unit tests
- `deployment/deploy.py` — reference Vertex AI Agent Engine deployment
- README with four worked example interactions
 
## Verification
 
- [x] `uv sync` clean
- [x] `uv run pytest tests/ -v` — 14 passed
- [x] All four Spraay tools import successfully
- [x] Agent loads in `adk web`
 
## Note on the dependency
 
The Spraay tools are merged into `google/adk-python-community` `main`
but have not yet shipped in a PyPI release. The `pyproject.toml` pins
`google-adk-community` to git `main` for now; this can be relaxed to a
standard version constraint once a release ships.
 
Apache 2.0 license matches the repo default.


plagtech:
Still active. Branch is updated against main, and all 8 findings from the automated reviews are addressed in 20378f2:

-Model ID bumped to gemini-3.5-flash per AGENTS.md (all 4 instances)
-Env-var handling now follows the recipe handbook: load_dotenv() in the package __init__.py, no hardcoded defaults in source, standard <TODO: update-this-value> placeholders in .env.example
-Weights are parsed to Decimal before summing in split_pool_proportionally
-The rounding-dust test now uses unequal weights and asserts exact per-recipient allocations
-Magic number replaced with a named ETH_ADDRESS_LENGTH constant

uv run pytest tests/ -v — 14 passed; agent loads in adk web. Happy to make any further changes — ready for review whenever.

happyhuman (commented):
Due to some restructuring of this repo, we no longer accept contribution in python/agents since all samples/recipes should now be placed under contrib/python. Please see [docs](https://github.com/google/adk-samples/tree/main/docs) for further information.

plagtech:
Thanks for the pointer! Moved same day — the recipe now lives at contrib/python/crypto-payroll-agent (via git mv, so history is preserved) and follows the recipe checklist: manifest.yaml, tests/test_runnability.py, aligned pyproject.toml, and uv run validate / ruff / uv lock all green.

I've also been addressing the automated review findings as they land each round. The security finding prompted the most substantive change: the spend ceiling, recipient cap, and address checks are now enforced programmatically in a before_tool_callback rather than living only in the instruction text — USD ceiling for pegged tokens, a separate ETH ceiling, and conservative blocking for tokens that can't be valued offline. The google-adk-community dependency is pinned to a commit SHA, and the full local test suite passes.

Happy to adjust anything else — ready for review whenever.

contrib/python/crypto-payroll-agent/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+Built on the [Spraay batch payment tools](https://github.com/google/adk-python-community/pull/95)
+in `google-adk-community`. Demonstrates a complete payroll, airdrop, and
+revenue-split workflow with built-in safety gates.
+
+## What this agent does
+
+A treasury operator says something like:
+
+> *"Pay 250 USDC each to alice.eth, bob.eth, and 0x742d…f44e."*
```
happyhuman: Can `alice.eth` work here? The safety gates section below says a recipient that fails 0x-format validation gets refused.

contrib/python/crypto-payroll-agent/README.md
```diff
@@ -0,0 +1,251 @@
...
+  print('Spraay tools ready')"
+```
+
+## Run the agent
+
+```bash
+uv run adk web
+```
+
+Then open `http://localhost:4200`, choose **crypto_payroll_agent**, and
```
happyhuman: Does `adk web` serve the dev UI on 4200? I had 8000 in mind.

contrib/python/crypto-payroll-agent/README.md
```diff
@@ -0,0 +1,251 @@
...
+batch.
+
+## Configuration
+
+| Variable | Required | Description |
+| --- | --- | --- |
+| `SPRAAY_PRIVATE_KEY` | yes | Private key of the sending wallet on Base |
+| `SPRAAY_RPC_URL` | no | Base RPC endpoint (default: `https://mainnet.base.org`) |
+| `SPRAAY_CONTRACT_ADDRESS` | no | Override the Spraay batch contract |
+| `GOOGLE_CLOUD_PROJECT` | yes¹ | GCP project for Vertex AI |
```
happyhuman, line 135: `GOOGLE_CLOUD_STAGING_BUCKET` has no row in this table, though `.env.example` declares it and `deploy.py` reads it with `os.environ[...]`. Can you please add it?

contrib/python/crypto-payroll-agent/README.md
```diff
@@ -0,0 +1,251 @@
...
+| `GOOGLE_CLOUD_LOCATION` | yes¹ | GCP region (e.g. `us-central1`) |
+| `GOOGLE_GENAI_USE_VERTEXAI` | yes¹ | `1` for Vertex AI, `0` for AI Studio |
+| `GOOGLE_API_KEY` | yes² | AI Studio key (alternative to Vertex AI) |
+| `PAYROLL_AGENT_MODEL` | no | Gemini model (default: `gemini-3.5-flash`) |
```
happyhuman: This row marks `PAYROLL_AGENT_MODEL` as not required, but `config.py` reads it through `_require_env`, which raises when it is unset. Which one is right?

contrib/python/crypto-payroll-agent/README.md
```diff
@@ -0,0 +1,251 @@
...
+The evalset covers all four batch-tool selection paths plus a safety-gate
+refusal case.
+
+## Testing
+
+```bash
+uv run pytest tests/ -v
+```
+
+All tests are offline. Spraay tools are mocked; helpers are tested
```
happyhuman: no mocks in any of the three test modules

contrib/python/crypto-payroll-agent/README.md
```diff
@@ -0,0 +1,251 @@
...
+directly.
+
+## Deployment
+
+See `deployment/deploy.py` for a reference Vertex AI Agent Engine
+deployment script.
+
+> **Note:** the script passes `PAYROLL_AGENT_MODEL` and
```
happyhuman: Should this mention `PAYROLL_MAX_BATCH_ETH` as well? `deploy.py` puts three entries in `env_vars`, not two.

contrib/python/crypto-payroll-agent/crypto_payroll_agent/prompt.py
```diff
@@ -0,0 +1,110 @@
...
+WORKFLOW
+
+1) UNDERSTAND
+   Identify: (a) ETH or ERC-20? (b) equal or variable amounts? (c) is the
+   user providing per-recipient amounts directly or a total + weights?
+
+2) RESOLVE
+   - If a token symbol is mentioned, call `lookup_token_info` first. Use
+     the returned address and decimals.
+   - If the resolution returns found=false, ASK the user for the
```
happyhuman: Step 2 tells the model to ask the user for a contract address when the lookup fails, but the safety gates further down refuse any token that is not USDC, USDbC or DAI. Is this branch reachable?

contrib/python/crypto-payroll-agent/crypto_payroll_agent/tools/helpers.py
```diff
@@ -0,0 +1,272 @@
...
+        "contract address and decimals before proceeding.",
+    )
+
+
+def split_pool_proportionally(
+    total_amount: str,
+    weights: list[float],
+    decimals: int = 6,
+) -> dict[str, Any]:
+    """Divide a token pool across recipients by integer weights.
```
happyhuman: the parameter three lines up is `list[float]`

contrib/python/crypto-payroll-agent/crypto_payroll_agent/guardrails.py
```diff
@@ -0,0 +1,180 @@
...
+        return False
+    if not value.lower().startswith("0x"):
+        return False
+    return all(c in "0123456789abcdefABCDEF" for c in value[2:])
+
+
+def _batch_total(tool_name: str, args: dict[str, Any], count: int) -> Decimal:
+    """Total amount a call would move, in the token's own human units.
+
+    Raises InvalidOperation if an amount is missing or unparseable, which
```
happyhuman: It seems a missing amount would come out of `args[equal_key]` as a `KeyError` rather than `InvalidOperation`. Does the caller catch both?

contrib/python/crypto-payroll-agent/pyproject.toml
```diff
@@ -0,0 +1,48 @@
...
+    "google-adk>=1.0.0",
+    # Spraay tools merged in adk-python-community#95 (May 2025) but not yet
+    # in a PyPI release. Pinned to an exact commit rather than `main` so the
+    # build is reproducible — tracking a moving branch means two installs a
+    # day apart can ship different tool code. The pinned commit is what
+    # uv.lock resolves and the tests run against; swap it for a version
+    # specifier once a PyPI release carries the Spraay tools.
+    "google-adk-community @ git+https://github.com/google/adk-python-community.git@396da17a9597d8d5f96e7e1aa8c1c396c738d146",
+    "google-cloud-aiplatform[adk,agent_engines]>=1.95.0",
+    "web3>=6.0",
```
happyhuman: nothing in the recipe imports `web3`

contrib/python/crypto-payroll-agent/deployment/deploy.py
```diff
@@ -0,0 +1,77 @@
...
+import tomllib
+import vertexai
+from vertexai.preview import reasoning_engines
+
+# Make the package importable from the repo root.
+sys.path.insert(
+    0, os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+)
+
+from crypto_payroll_agent import root_agent
```
happyhuman, line 36: These two imports sit below the `sys.path.insert`, so ruff will raise E402 on them. You may add `# noqa: E402`, the same way `crypto_payroll_agent/__init__.py` does for its post-bootstrap import.

contrib/python/crypto-payroll-agent/eval/crypto_payroll_eval_set.evalset.json
```diff
@@ -0,0 +1,173 @@
+{
+  "eval_set_id": "crypto_payroll_eval_set_001",
+  "name": "Crypto Payroll Agent — tool selection + safety",
+  "description": "Verifies the agent selects the correct Spraay batch tool for equal-vs-variable and ETH-vs-token, calls lookup_token_info for symbols, and refuses out-of-bounds requests.",
```
happyhuman, line 4: The description claims equal-vs-variable and ETH-vs-token, but none of the five cases calls `spraay_batch_eth_variable`. Worth adding one?

contrib/python/crypto-payroll-agent/tests/test_agents.py
```diff
@@ -0,0 +1,247 @@
...
+    # still distributes to the cent.
+    result = split_pool_proportionally("100.123456", [1, 1], decimals=6)
+    assert result["ok"] is True
+    assert Decimal(result["total_distributed"]) == Decimal("100.123456")
+
+
+def test_split_pool_accepts_trailing_zeros_beyond_decimals():
+    from crypto_payroll_agent.tools.helpers import split_pool_proportionally
+
+    # "100.0000000" has 7 fractional digits but only 2 significant ones
```
happyhuman: `100.0000000` has no significant digits past the point, not 2

contrib/python/crypto-payroll-agent/tests/test_agents.py
```diff
@@ -0,0 +1,247 @@
...
+    # Spraay batch tools
+    assert "spraay_batch_eth" in tool_names
+    assert "spraay_batch_token" in tool_names
+    assert "spraay_batch_eth_variable" in tool_names
+    assert "spraay_batch_token_variable" in tool_names
+
+    assert root_agent.name == "crypto_payroll_agent"
+
+
+if __name__ == "__main__":
```
happyhuman: the other two test modules don't carry this block

contrib/python/crypto-payroll-agent/crypto_payroll_agent/config.py
```diff
@@ -0,0 +1,108 @@
+"""Configuration for the Crypto Payroll Agent."""
```
happyhuman, line 15: No licence header on this one, and none of the 12 `.py` files in the recipe has one. I suggest adding the standard Apache header; you may borrow it from a recipe under `core/`.

contrib/python/crypto-payroll-agent/pyproject.toml
```diff
@@ -0,0 +1,48 @@
...
+license = { text = "Apache-2.0" }
+dependencies = [
+    "google-adk>=1.0.0",
+    # Spraay tools merged in adk-python-community#95 (May 2025) but not yet
+    # in a PyPI release. Pinned to an exact commit rather than `main` so the
+    # build is reproducible — tracking a moving branch means two installs a
+    # day apart can ship different tool code. The pinned commit is what
+    # uv.lock resolves and the tests run against; swap it for a version
+    # specifier once a PyPI release carries the Spraay tools.
+    "google-adk-community @ git+https://github.com/google/adk-python-community.git@396da17a9597d8d5f96e7e1aa8c1c396c738d146",
```
happyhuman: Nice, pinning the exact commit with the reason spelled out beats tracking `main`.  :-)

happyhuman:
One more, on a line the diff view won't let me anchor to:

- `uv.lock:1091` — `google-adk-community` resolves from a git source, and the recipe dependency policy only allows published PyPI releases. This one is a hard CI failure rather than a nit, so it will need sorting before the PR can go in.

happyhuman (commented):
Thanks for moving your recipe. I left a few more comments. PTAL.

contrib/python/crypto-payroll-agent/README.md
plagtech: Good catch — swapped to 0x example addresses and noted ENS isn't supported.

contrib/python/crypto-payroll-agent/README.md
plagtech: You're right, 8000 — verified locally, fixed.

contrib/python/crypto-payroll-agent/deployment/deploy.py
plagtech: Turns out ruff's E402 exempts imports that follow a sys.path modification, so the directive would be flagged unused (RUF100) under the root config — left as-is; the __init__.py noqa survives because it follows load_dotenv(), not a path edit.

contrib/python/crypto-payroll-agent/eval/crypto_payroll_eval_set.evalset.json
plagtech: Added a sixth case covering spraay_batch_eth_variable.

plagtech:
Thanks for the thorough review — all comments addressed in 2ccfb3c. Answers to the specific questions, and the blocker:

uv.lock / git source (the CI blocker): Resolved. The dependency is now google-adk-community[spraay]>=0.5.0 from PyPI — the spraay extra carries web3 (a bare install leaves it absent and the tools fail on import), and the lock resolves from the registry with sha256 hashes. I verified the 0.5.0 wheel's spraay module is byte-identical to the commit that was pinned, so this is purely a sourcing change. This also answers the web3 comment: it's gone from the recipe's own deps and arrives via the extra.

alice.eth: Good catch — the example now uses 0x addresses, and the README notes ENS isn't supported.

adk web port: You're right, it's 8000 (verified locally). Fixed.

PAYROLL_AGENT_MODEL required?: The table was stale — it's required (_require_env raises); the table now says so, with the default noted as coming from .env.example.

Unreachable prompt branch: Fixed in two places — the prompt and lookup_token_info's note field (which the model reads verbatim) both now explain that non-pegged tokens aren't supported offline, instead of soliciting an address.

KeyError vs InvalidOperation: Right — a missing amounts argument now returns a clean refusal naming the argument, the docstring is corrected, and tests cover both the missing and non-numeric paths.

E402 noqa: Interesting one — ruff's E402 exempts imports that follow a sys.path modification, so the directive comes back as RUF100 (unused) under the root config. Left the code as-is; the existing noqa in __init__.py survives because it follows load_dotenv(), not a path edit.

Eval coverage: Added a sixth case exercising spraay_batch_eth_variable.

The rest (staging-bucket table row, deploy note now listing three env vars, "mocked" wording, integer-weights docstring, __main__ block, Apache headers on all the .py files) are applied as suggested. 35 tests passing locally. PTAL whenever convenient — happy to adjust anything further.

PR #1978, opened 2026-05-23
livekit integration with adk with gemini live

kishorerj:
This is a sample application illustrating how livekit and ADK can work together to create a complete voice based applications with Gemini Live API. Livekit is one of the framewroks recommended by Google for use with Gemini Live, however no regference archiotecture or sample code base exist. In this livekit framework provides stable WebRTC connection for the end user, while ADK framework is a fist class agentic framework to define agents, maintain conversation history etc. This gives a working architecture for customers or users to use Gemini Live in Production .


kishorerj:
@happyhuman @eliasecchig @mstyer-google Please review and let me know if there are any queries

PR #2168, opened 2026-07-01
feat: add recipe validation tools

happyhuman:
## Summary

- Adds a `tools/` directory with a self-contained uv project for local recipe validation
- Provides a single `validate` entry point with subcommands for running individual or all checks
- Contributors can validate recipes locally before submitting a PR, catching issues early

## Usage

```bash
cd tools && uv sync
uv run validate manifest core/rag-agent-search
uv run validate all
```

## Structure

- `tools/validate.py` — dispatcher entry point, routes subcommands to individual tools
- `tools/validate_manifest.py` — checks manifest.yaml presence and schema conformance
- `tools/pyproject.toml` — self-contained uv project with dependencies
- `tools/README.md` — usage instructions


eliasecchig (changes requested):
You will need to refresh the lock forcing pipy


PS we could also add it as separate Github action:

Add 
[.github/workflows/validate-lockfiles.yml](http://localhost:3100/ELI/issues/ELI-12?file=.github%2Fworkflows%2Fvalidate-lockfiles.yml)
 that scans all uv.lock files in the repo and fails CI if any source.registry or package URL contains pkg.dev or other non-PyPI hosts. A bash/grep check is enough. Trigger on PRs touching *.lock files, push to main, and workflow_dispatch. Use SHA-pinned actions per repo convention. This prevents the issue from silently slipping through in future PRs.

happyhuman:
> You will need to refresh the lock forcing pipy
> 
> PS we could also add it as separate Github action:
> 
> Add [.github/workflows/validate-lockfiles.yml](http://localhost:3100/ELI/issues/ELI-12?file=.github%2Fworkflows%2Fvalidate-lockfiles.yml) that scans all uv.lock files in the repo and fails CI if any source.registry or package URL contains pkg.dev or other non-PyPI hosts. A bash/grep check is enough. Trigger on PRs touching *.lock files, push to main, and workflow_dispatch. Use SHA-pinned actions per repo convention. This prevents the issue from silently slipping through in future PRs.

I like the idea of adding the lockfile validator as a workflow. But the link above points to a localhost file, so I am unable to open it. As a matter of fact, would you like to add it to the repo yourself and send me the PR? If not, please share the file with me with a correct link and I will add it. We can merge this PR in that case.

eliasecchig (approved):

eliasecchig (approved):

PR #2183, opened 2026-07-06
refactor: migrate all agents to Go ADK v2

hanorik:


karolpiotrowicz (approved):

karolpiotrowicz:
I'd want to close this because the repository was reorganized underneath it, not because of anything wrong with the change.

Two things moved since July. `boat-agent`, `navallist` and `sail-researcher` were removed in #2291, which is where all the conflicts here come from — those three no longer exist on main. And `go/agents/` has since been retired in favor of `contrib/go/`, so pull requests that modify files in place under it no longer pass CI. The [go/README.md](https://github.com/google/adk-samples/blob/main/go/README.md) notice covers the new layout.

The v2 migration is still wanted for the two agents that remain. `financial-advisor` is pinned to `google.golang.org/adk v0.6.0` and `llm-auditor` has no `go.mod` at all, so both need exactly what you did here. The route for it now is a move to `contrib/go/<name>`, which the retired-folder rule exempts because it is a rename rather than an in-place edit, following the [recipe checklist](https://github.com/google/adk-samples/blob/main/docs/recipe-checklist.md). Your changes to those two agents carry over nearly unchanged, and I would be glad to review that one.

Reopening is one click if you would rather continue here, and your commits are untouched on your branch either way.

PR #2185, opened 2026-07-06
fix: remove trailing slashes from integration test BASE_URLs to prevent double-slash

jackwotherspoon:
Slash at end of BASE_URL causes double-slash in session URLs which results in 404 for integration tests.

Double-url causing line:
```python
session_url = f"{BASE_URL}/apps/software_bug_assistant/users/{user_id}/sessions"`
```


jackwotherspoon:
@eliasecchig mind taking a quick look at this one 😄

jackwotherspoon:
Actually this may be a bug in `agent-starter-pack`...

PR #2188, opened 2026-07-06
chore: reduce noise in Gemini PR review workflows

happyhuman:
## Summary

The three Gemini PR review workflows (security, correctness, maintainability) were generating too much noise: multiple long structured comments per PR plus three separate review-triggered emails to the PR author on every run.

## Changes

- **No more emails:** switched from the pending review flow (`create` → `add` → `submit`) to direct inline comments via `create_pull_request_review_comment`. Formal review submission is what triggers GitHub's review notification emails; removing it eliminates that.
- **High/critical only:** added an explicit severity filter to each prompt. The agent now silently skips `medium` and `low` findings entirely and posts nothing if there are no qualifying issues.
- **Shorter, plainer comments:** replaced the structured `<COMMENT>` block format (emoji severity header + markdown suggestion block) with an instruction to write 1–2 plain sentences — what's wrong and how to fix it — like a colleague leaving a quick note.

The maintainability reviewer is unchanged in scope but will now rarely comment in practice, since its issues are `low`/`medium` by nature. That's expected and fine.


ToniCorinne (approved):
Changes lgtm, but I am curious to the reasoning for removing medium severity issue feedback. Wouldn't we still want the author to be aware of those issues and then be able to make the judgement call about fixes? Does that affect noise levels that drastically?

PR #2301, opened 2026-07-26
feat(long-horizon-harness): add the Horizon long-horizon agent recipe

eliasecchig:
## What this adds

`core/python/long-horizon-harness` — **Horizon**, a self-improving long-horizon agent on ADK + Vertex AI.

Most recipes here show a pattern in one file. This one is deliberately the opposite: a complete system for the problems that only appear once an agent runs for **hours instead of turns** — keeping context bounded, giving tools a real machine to work on, and letting the agent improve at its own job over time.

## What it demonstrates

| Interface | What it shows |
|---|---|
| **Sandboxed environment** | Every tool call goes through an `Environment` ContextVar — a local backend for tests, a per-user Agent Runtime BYOC container in deploy — instead of touching the host. Hermetic by default when deployed. |
| **Self-improvement on Memory Bank** | A per-turn judge fork, a pre-compaction flush, and a nightly dream-review pass that consolidates a native Structured Profile — layered on the managed store, no custom vector DB. |
| **Layered tool guardrails** | Exfiltration, policy, and an interactive permission ask-layer composed as an ordered callback chain. |
| **Delegation with HITL resurfacing** | A sub-agent can bubble one approval — or a free-text question — up to the user mid-run, then resume in place. |
| **Routines** | Unattended recurring tasks in isolated, secret-scoped sandboxes. |
| **A2A + web client** | A2A surface plus a Vite + TanStack React client. |

Custom code is confined to those interfaces. Compaction, resumability, and prefix caching are ADK/Vertex knobs the recipe only *configures* — the README and `AGENTS.md` are explicit about that boundary, so readers can tell what they'd have to build versus what the platform already gives them.

## Validation

```
uv run validate core/python/long-horizon-harness
  [PASS] Manifest validation
  [PASS] Structure validation
  [PASS] README validation

cd core/python/long-horizon-harness
uv run pytest --ignore=tests/integration --ignore-glob="**/test_integration.py"
  2274 passed, 4 skipped
```

`check_lockfile_hashes.py` passes. `check_recipe_pyproject.py` passes on `requires-python`, `description`-matches-manifest, and the public-PyPI `[[tool.uv.index]]`.

## Why `large: true`

790 files, ~7.8 MB — past `core/`'s 500-file default tier, well inside the `large` tier. The size *is* the point: a long-horizon harness that omitted the sandbox, memory loop, or guardrails wouldn't demonstrate the thing it exists to demonstrate.

## Open items — feedback welcome before I take this out of draft

Three `python-validate-recipe` rules conflict with the upstream source repo this is mirrored from, and I'd rather agree on the direction than guess:

1. **`[project].name` is `lha`, not the folder name.** The name is a deploy contract upstream — `agents-cli deploy` derives the Cloud Run service name from it, so renaming repoints a live service. Happy to rename here and carry the divergence, if that's preferred over an exemption.
2. **The recipe keeps a `[tool.ruff]` block.** Removing it (per the A1 rule) puts the code under the root config: **304 files reformat** (88 → 80 cols) and 365 lint findings, mostly `PL` rules and isort treating `horizon` as third-party. If the root `known-first-party` could take `horizon` alongside `app`, most of the isort churn disappears — I'd be glad to send that one-liner separately.
3. **36 env vars are declared as comments in `.env.example`.** The checker only counts uncommented lines, but uncommenting them makes the sample's own `make dev` seed a `.env` full of empty values that override real defaults (`LHA_SANDBOX_LOCATION=""` instead of `us-central1`). I'd like to add them with real default values rather than blanks — confirming that's acceptable.

Draft until those three are settled.


eliasecchig:
Superseded — reopening from a branch on this repo directly rather than a fork.

PR #2492, opened 2026-08-07
Editable documents in the side panel

lspataroG:
# Editable documents in the side panel

The side panel used to be read-only: you could look at a file the agent wrote, but not change it. Now it's a place you can work — open a document, edit it, and point the agent at a sentence.

46 files · no new dependencies

---

## What you can do now

**Open any file by clicking it** in the sidebar. Before, clicking only downloaded it.

**Edit documents in the panel.** Type directly into the file. It saves itself a moment after you stop typing — no save button. A file whose bytes aren't valid UTF-8 stays read-only rather than being saved back mangled.

**Format with a toolbar**, like a document editor: bold, italic, headings, bullet and numbered lists, quotes, links, tables, code.

**Highlight a sentence and tell the agent what to do with it.** Select some text, a small chip appears above the message box, then type *"make this shorter"*. The agent changes exactly that sentence and nothing else — even in a file where the same words appear several times. This works while the agent is busy too: the message waits its turn and keeps the highlight.

**The panel keeps itself up to date.** When the agent creates a file, it opens. When the agent edits the file you're looking at, it refreshes. When a file is renamed or deleted, the tab follows. No reloading the page.

**Click a filename in the chat to open it.** Whether the agent writes it as a link or just mentions it, the name is clickable.

**Drag a file from the sidebar into the message box** to attach it, instead of typing its name.

**The agent never fights you for the panel.** While it's working, the editor locks so your changes can't be overwritten. Its own scratch files (plans, helper scripts) stay out of your way.

---

## How it works, briefly

The file on disk stays plain Markdown, HTML or text — nothing proprietary, and the agent can still read and edit it normally.

When you highlight text, the panel records **where** it is in the file rather than **what** it says. That's what makes editing reliable in a document that repeats itself, and it's also what would let editing move into the rendered preview later.

UI actions — a highlight, a dragged file — are sent to the agent as structured events, not as text pasted into your message. So the chat shows what you did, not the machinery behind it.

---

## Technical summary

| Change | Purpose |
|---|---|
| `PUT /lha/workspace/file` | Save an edit from the browser. Same guards as the agent's own writes: protected paths, binary reject, traversal, size cap. |
| `patch` accepts character offsets | Edit an exact span. The expected text is verified before writing, so a stale edit fails rather than corrupting the file. |
| `datapart.py` renders UI events | Turns a highlight or an attached file into instructions server-side, so none of it appears in the transcript. |

## Testing

- Web **492 passing**, Python **2411 passing**
- 6 opt-in end-to-end tests (`SMOKE_LIVE=1`) that drive the real agent: editing a highlighted passage, doing it in a repetitive file, panel refresh, auto-open, rename, HTML editing

**Pre-existing failures, untouched:** 11 in `use-appearance.test.ts` (a happy-dom issue), 1 in `test_repo_overview.py`. Both fail on `main`.

---

## Limits

- **The rendered preview is read-only.** Agent-generated HTML runs in a locked sandbox, so the page can't read clicks or selections inside it. HTML is edited as source.
- **Fonts and colours aren't available.** Markdown has no syntax for them. Headings give you six text sizes.

### DIFF
--- a/core/python/long-horizon-harness/horizon/a2a/datapart.py
+++ b/core/python/long-horizon-harness/horizon/a2a/datapart.py
@@ -33,6 +33,72 @@ import json
 A2A_DATA_PART_START_TAG = b"<a2a_datapart_json>"
 A2A_DATA_PART_END_TAG = b"</a2a_datapart_json>"
 
+# Discriminator the web client stamps on a highlighted-selection payload.
+# Mirrors SELECTION_DATA_KIND in web/components/chat/workspace-href.ts.
+_SELECTION_KIND = "lha.selection"
+
+# Discriminator for files dragged from the workspace tree into the composer.
+# Mirrors FILEREF_DATA_KIND in web/components/chat/workspace-href.ts.
+_FILEREF_KIND = "lha.fileref"
+
+
+def _render_fileref(payload: dict) -> str | None:
+    """Prose for workspace files the user attached from the tree.
+
+    Only paths travel: the files are already in the workspace, so re-sending
+    their bytes would duplicate them. Says what was pointed at, not what to do
+    with it — the user's own message carries the instruction.
+    """
+    paths = payload.get("paths")
+    if not isinstance(paths, list) or not paths:
+        return None
+    clean = [p for p in paths if isinstance(p, str) and p]
+    if len(clean) != len(paths):
+        return None
+    listed = "\n".join(f"- {p}" for p in clean)
+    noun = "file" if len(clean) == 1 else "files"
+    return (
+        f"[The user attached {len(clean)} workspace {noun} to this message. "
+        "Read them if the request concerns them:]\n"
+        f"{listed}"
+    )
+
+
+def _render_selection(payload: dict) -> str | None:
+    """Prose for a text selection the user highlighted in the web UI.
+
+    Hands over character offsets rather than asking the model to locate the
+    text: the same words may appear many times in a file, and a search anchor
+    cannot tell which one was pointed at. Deliberately states what was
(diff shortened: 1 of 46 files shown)


lspataroG:
Rebased onto latest `main` and squashed to a single commit.

### Note on the failing `Ruff Lint Suggestions` check

That failure looks unrelated to this PR — I think the workflow fails on **any** PR touching Python right now.

`python-format.yml` decides the job from whether ruff printed anything:

```bash
LINT_OUTPUT=$(uvx ruff check --force-exclude --output-format=concise --exit-zero "${CHANGED_FILES[@]}")
if [ -n "$LINT_OUTPUT" ]; then ... exit 1
```

`uvx ruff` is unpinned, and as of **ruff 0.16.1** a clean run prints `All checks passed!` to stdout in `concise` format. So `LINT_OUTPUT` is non-empty even with zero violations, and the job exits 1.

The run log for this PR shows exactly that — the "Ruff lint violations" group contains only:

```
All checks passed!
```

Reproduced locally on the files in this PR:

```console
$ uvx ruff check --force-exclude --output-format=concise --exit-zero <changed files>
All checks passed!
$ uvx ruff --version
ruff 0.16.1
```

A one-line fix would be to filter that sentinel out, e.g.:

```bash
LINT_OUTPUT=$(uvx ruff check ... --exit-zero "${CHANGED_FILES[@]}" | grep -v '^All checks passed!$' || true)
```

Happy to open a separate PR for it if that's useful — it didn't feel right to change CI inside a feature PR.

eliasecchig (approved):

PR #2495, opened 2026-08-10
fix(small-business-loan-agent): empty-string enum schema bug and LLM-judge multi-turn false-block

ProDost:
## Summary

Two bugs found while running `small-business-loan-agent` end-to-end against the Gemini API key surface:

1. **`loan_purpose` empty-string enum bug.** `LoanApplicationData.loan_purpose` is a `Literal[...]` that includes `""` as a member, so Pydantic's generated JSON schema puts an empty string in the enum. The current `google-genai` structured-output validation rejects that outright:
   ```
   400 INVALID_ARGUMENT: GenerateContentRequest.generation_config.response_schema
   .properties[loan_purpose].enum[6]: cannot be empty
   ```
   This makes `DocumentExtractionAgent` fail on every single request — the sample cannot process any document as shipped. Fixed by making the field `Optional[Literal[...]]` with `default=None` instead of including `""` as a literal value. No behavior change downstream — the missing-field checks in `state_callbacks.py` treat `None` the same way they treated `""`.

2. **LLM-judge gate false-blocks on any multi-turn continuation.** `llm_judge_gate`'s trajectory check (`_extract_tool_sequence_and_messages`) only inspects tool calls made in the *current* invocation (`session.events` filtered by `invocation_id`). Since a loan application is processed across multiple user turns by design (submit → approve, or submit → pause-for-repair → resume), any turn that correctly skips a step completed in an earlier turn gets misjudged as "skipped mandatory steps," and the judge blocks the response — even though the underlying decision was correctly made and persisted to Firestore. I reproduced this on both:
   - the post-approval turn (only `LoanDecisionAgent` is called this turn — Extraction/Underwriting/Pricing ran in the prior turn)
   - the post-repair resume turn (skips the already-completed `DocumentExtractionAgent`)

   Fixed by adding `_build_process_history()`, which reads the persisted `ProcessStateService` state to reconstruct which `ALL_STEPS` entries were already completed *before* this turn, and passes that to the judge prompt as an explicit "Process History" section, distinct from "Tool Call Sequence (this turn only)". The judge prompt now explains the multi-turn nature of the workflow and only treats a step as genuinely missing when Process History confirms it wasn't done in any prior turn either.

## Verification

Ran the full documented flow (happy path with approval, and pause/repair/resume) end-to-end against the Gemini API with real model calls:
- Fix 1: `DocumentExtractionAgent` now successfully returns structured output instead of a 400 on every call.
- Fix 2: both previously-blocking scenarios (approval-only turn, resume-after-repair turn) now get `trajectory_correct=True, grounded_in_context=True` from the judge, with no regression on the single-turn full-pipeline case (extraction → underwriting → pricing all in one turn, which already passed before this change).

## Notes for reviewers

- This PR is intentionally scoped to just these two bugs. I initially adapted this sample to run locally against a plain Gemini API key (instead of Vertex AI) and a local JSON state store (instead of Firestore) to get it running without GCP access — those changes are local-dev-only workarounds and are **not** included here, since they'd replace the sample's intended Vertex/Firestore architecture rather than fix a bug in it.
- I have not yet signed the Google CLA — will do so via the CLA bot on this PR.


happyhuman:
We have recently introduced a major restructuring of this repo and we no longer accept receipes under python/agents. Please take a look at the docs to learn more about the new changes and how you can update your code properly and resubmit it under contrib/python:
https://github.com/google/adk-samples/tree/main/docs

happyhuman (commented):
Please move this recipe to contrib/python before submitting this PR. See [docs/](https://github.com/google/adk-samples/tree/main/docs) for more info.

PR #2506, opened 2026-08-12
feat(contrib): cross-border data-policy router

janrockdev:
## Summary

Adds `contrib/python/cross-border-data-router`, a Python ADK recipe demonstrating a common enterprise compliance problem: **deciding which regional agent is legally allowed to process a piece of data, based on where that agent says it operates.**

A root orchestrator receives a data-processing request (a classification like `PII` plus an origin region), evaluates it against a registry of regional processor agents, and only then delegates the work — or refuses outright if no agent qualifies. Eligibility is read directly off each candidate's real `a2a.types.AgentCard` via a custom `capabilities.extensions` entry, since the core A2A spec has no built-in notion of jurisdiction or data residency.

## Why this recipe

I searched the repo first for an existing sample that evaluates policy and routes based on an agent's declared location/jurisdiction, and found none:

- The only prior A2A `AgentCard` usage in the repo (`core/python/long-horizon-harness`) builds/serves a card but never reads location off one.
- `python/agents/global-kyc-agent` routes UK vs. US sub-agents, but by an LLM reading the *company being checked* out of free text — not metadata declared by the candidate agent itself.
- "Policy" elsewhere in the repo (`long-horizon-harness`'s tool-call guardrails, `core/python/safety-plugins`) means something different: gating a single agent's own actions, not choosing between agents.

This recipe fills that gap with a pattern modeled on [OpenEAGO](https://openeago.finos.org/) ([repo](https://github.com/finos-labs/open-eago)), a FINOS Labs specification for jurisdiction-aware governance of multi-agent workflows. OpenEAGO isn't a published Python package, so this is a from-scratch implementation of the pattern it describes (two-stage hard-filter-then-score agent selection), not a dependency on it — credited and linked in the README.

## What's in it

- `app/policy/cards.py` — a static registry of three regional processor cards (`eu_processor`, `uk_processor`, `us_processor`), each a real `a2a.types.AgentCard` carrying jurisdiction/data-residency metadata via an `AgentExtension`.
- `app/policy/engine.py` — the routing policy: Stage 1/2 hard filters (data residency, jurisdiction exclusion) that scoring can never override, then a weighted score among survivors. No candidates surviving means an explicit rejection, never a silent fallback to a non-compliant agent.
- `app/agent.py` + `app/sub_agents/*` — an ADK `LlmAgent` orchestrator that calls the policy tool first, then delegates to the approved regional sub-agent via `AgentTool`.
- `tests/unit/test_policy_engine.py`, `tests/unit/test_tools.py` — cover the hard filters, preference scoring, and the reject-rather-than-fallback path directly.

## Testing

Run locally from the repo root:

```bash
export RECIPE_PATH=contrib/python/cross-border-data-router
uv run validate "$RECIPE_PATH"                # PASS — manifest + structure
uv run ruff format "$RECIPE_PATH" && uv run ruff check "$RECIPE_PATH"   # clean
cd "$RECIPE_PATH" && uv lock && uv run pytest --ignore=tests/integration --ignore-glob="**/test_integration.py"
# 11 passed
```

Also manually verified `uv run adk run app` and `uv run uvicorn app.fast_api_app:app --reload` both start and route correctly against a live Gemini API key.

### DIFF
--- a/contrib/python/cross-border-data-router/.env.example
+++ b/contrib/python/cross-border-data-router/.env.example
@@ -1,0 +1,27 @@
+# Model Configuration
+MODEL_NAME=gemini-3.5-flash
+
+# Google Cloud Platform Configuration (for Vertex AI)
+# GOOGLE_CLOUD_PROJECT=<TODO: update-this-value>
+# GOOGLE_CLOUD_LOCATION=global
+# GOOGLE_GENAI_USE_VERTEXAI=True
+
+# Google AI Studio Configuration (if using API Key instead of Vertex AI)
+# GEMINI_API_KEY=<TODO: update-this-value>
+
+# Web Server and Telemetry Configuration (all optional; see app/fast_api_app.py
+# and app/app_utils/telemetry.py). Leave commented out for local development —
+# uncomment and fill in only if you're deploying with GCS-backed logging/tracing.
+# Setting LOGS_BUCKET_NAME to anything other than a real bucket name (e.g. a
+# TODO placeholder) will break `uv run uvicorn app.fast_api_app:app`, since a
+# non-empty value is treated as "GCS artifact storage is enabled."
+# ALLOW_ORIGINS=http://localhost:3000,http://localhost:8080
+# LOGS_BUCKET_NAME=your-gcs-bucket-name
+# OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=NO_CONTENT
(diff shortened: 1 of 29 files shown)


janrockdev:
CLA added.

janrockdev:
recheck

janrockdev:
The CLA blocker is resolved — signed and confirmed. Triggered a recheck on the `cla/google` status check above; it may take a few minutes to flip green.

@happyhuman — ready for review whenever you have a chance. Let me know if anything needs changes.

PR #2538, opened 2026-08-18
docs(contrib): fix dead handbook link and extend the link check to the entry-point READMEs

joshnaim1:
## What's broken

`contrib/README.md` links to `../docs/recipe-handbook/folders/contrib.md`, which
404s. There is no `folders/` directory under `docs/recipe-handbook/`, and
`git log -- docs/recipe-handbook/folders/contrib.md` returns no commits, so the
page was never created. The link came in with the handbook in #2277 and has
pointed at nothing since.

The documentation it should reach already exists. `docs/recipe-checklist.md` and
`docs/recipe-handbook/README.md` both cover contributing to `contrib/`. Only the
path is wrong.

## The fix

Point it at `docs/recipe-checklist.md`, which opens with "Everything you need to
submit a `contrib/` recipe, on one page" — the requirements and shipping
checklist the sentence promises. The handbook README stays as the secondary link
for deeper detail. That ordering matches how `CONTRIBUTING.md` and
`docs/README.md` already route contributors.

## Why CI missed it

`docs-links.yml` runs lychee and fails on dead relative paths, but both its
trigger paths and its scan glob are `docs/**`. `contrib/README.md` sits outside
both, so the dead link never got checked.

The second commit adds the four entry-point READMEs (`README.md`,
`CONTRIBUTING.md`, `contrib/README.md`, `core/README.md`) to the trigger paths
and the glob. These are the files a new contributor reads first and they all
link into `docs/`.

I kept the scope deliberately narrow. Those four files have 11 relative links
between them and the only broken one is the one fixed here, so the check goes
green. Widening the glob repo-wide would fail immediately — there are ~29 other
dangling links under `python/agents/**` and `kotlin/README.md`. That is older,
separate rot and belongs in its own PR.

## Testing

- Confirmed the target path 404s on `raw.githubusercontent.com` and is absent
  from the repo tree.
- Scanned all 205 markdown files for dangling relative links to check this was
  not part of a wider pattern. It is the only `folders/` reference in the repo.
- Verified the four newly covered files contain no other broken relative links,
  so the widened check passes on this branch.


.github/workflows/docs-links.yml
```diff
@@ -29,11 +42,11 @@ concurrency:
 
 jobs:
   validate-docs-links:
-    name: Check every link in docs/ resolves
+    name: Check every documentation link resolves
```
happyhuman: Renaming the job renames the status check. Is `Check every link in docs/ resolves` still required on `main`?

contrib/README.md
```diff
@@ -18,4 +18,5 @@ with setup and run instructions.
 ## Contributing to `contrib/`
 
 For the requirements and shipping checklist, see the
-[Recipe Handbook — `contrib/` page](../docs/recipe-handbook/folders/contrib.md).
+[recipe checklist](../docs/recipe-checklist.md). For deeper
```
happyhuman, line 21: lowercase "recipe checklist" but title-case "Recipe Handbook"

happyhuman (dismissed):

.github/workflows/docs-links.yml
joshnaim1: Good catch, I hadn't thought about the status check name being a contract. I can't read branch protection from outside the org, so I couldn't tell whether that string is required on main. Reverted the rename rather than guess. The widened scope was the actual point, the name was just cosmetic. If you want the check renamed later that's easy to do from your side.

contrib/README.md
joshnaim1: Fixed. Lowercased both so they match, following the same style CONTRIBUTING.md uses.

happyhuman (approved):

PR #2472, opened 2026-08-05
feat(skills): add retail/virtual-tryon

tanvisinghal-0105:
## Summary

Adds `skills/retail/virtual-tryon/` — a conversational installer skill that stands up a virtual try-on agent on Google Cloud (Vertex AI image + Veo video). First contribution under the `skills/retail/` vertical.

- **What it does:** Guided setup for a VTO fitting-room sandbox: Q-MODE interview → creates GCS buckets → verifies Vertex AI model access → launches a local FastAPI sandbox at `localhost:8080` with a demo catalog (shirt + sunglasses). Also supports "export to Cloud Run" mode that generates a standalone containerized deployment.
- **What's new to teach:** How to compose Gemini image models (`gemini-2.5-flash-image` / `gemini-2.5-pro-image`) with Veo 3.1 reference-to-video for a two-stage image → video try-on pipeline. Also demonstrates the ADK `agents.Agent` + `apps.App` pattern in a `scripts/` layout (as opposed to the `app/` layout).

## Validator status

Ran `uv run validate skills/retail/virtual-tryon` from the repo root:

| Check | Result |
|---|---|
| Manifest validation | PASS |
| Structure validation | **FAIL — `.env.example` missing** (see below) |
| README validation | PASS |
| Placement validation | PASS |

Also ran:
- `uv run ruff format --check skills/retail/virtual-tryon` → PASS
- `uv run ruff check skills/retail/virtual-tryon` → PASS
- `cd skills/retail/virtual-tryon && uv sync --dev && uv run pytest --ignore=tests/integration` → **1 passed**

## About the `.env.example` failure

This skill's primary configuration channel is **`design-spec.md`**, not `.env`. When a user invokes the skill (via `/retail-virtual-tryon` in Claude Code or "Use the retail-virtual-tryon skill …" in Antigravity), Q-MODE writes user answers into `design-spec.md`, and `scripts/start_sandbox.py:apply_config_to_env` exports those values into `os.environ` before booting the FastAPI server. `.env` never participates in the runtime path.

Adding a placeholder `.env.example` would satisfy the file-existence check but would be misleading: it would suggest a config channel the skill doesn't actually use. 
## Test plan

- [ ] `uv run validate skills/retail/virtual-tryon` — matches the results above (structure fail on `.env.example` only)
- [ ] `uv run ruff format --check skills/retail/virtual-tryon` — no changes suggested
- [ ] `uv run ruff check skills/retail/virtual-tryon` — no errors
[...]

PR #2555, opened 2026-08-24
fix(virtual-tryon): unbreak the Cloud Run export and correct the prerequisites

gabrielahrlr:
## Summary

Two follow-ups to #2472, which merged on Aug 21.

**`fix`: the exported Cloud Run app could not start.** `scripts/export_app.py`
copied five modules into the generated bundle but not `config.py`, which every
one of them imports — so the rewritten `from config import config` resolved to
nothing. Fixing that surfaced a second crash: `config.py` calls `load_dotenv()`
and `python-dotenv` was missing from the export template's `requirements.txt`.

**`docs`: the README stated prerequisites that do not hold.** It required
`retail/product-search` and claimed that skill's Vector Search collection
supplies this one's catalog. Neither is true — nothing under `scripts/`
references it, and the recipe ships its own catalog. `EVAL.yaml` graded the
agent on repeating the claim, so that rubric line goes too; removing only the
README text would have failed the eval. Also adopts the Gemini Enterprise Agent
Platform name in user-facing text (Python comments and the manifest's
`vertex-ai` service slug are left as-is, the slug being a machine identifier
shared with other recipes), and documents the Cloud Run export path.

## Test plan

- [x] `uv run validate skills/retail/virtual-tryon` — manifest, structure, README and placement all PASS
- [x] `uv run ruff format --check skills/retail/virtual-tryon` — 11 files already formatted
- [x] `uv run ruff check skills/retail/virtual-tryon` — all checks passed
- [x] `cd skills/retail/virtual-tryon && uv run pytest --ignore=tests/integration` — 1 passed
- [x] Cloud Run export end to end: exported the app, installed the generated `requirements.txt` into a clean venv, and ran the Dockerfile's own `uvicorn server:app`


skills/retail/virtual-tryon/assets/export-template/requirements.txt
happyhuman, line 1: We don't generally want the recipes to use requirements.txt. Is there a reason you need to use one here?

gabrielahrlr: I've fixed the underlying problem.

For context on the original: this wasn't the recipe's dependency file. The recipe uses `pyproject.toml` + `uv.lock`. That `requirements.txt` lived in `assets/export-template/` and was a template for the *generated* standalone Cloud Run bundle, consumed by the generated Dockerfile at image build. It came in with #2472; this PR was only adding the missing `python-dotenv` line.

happyhuman (approved):

PR #2564, opened 2026-08-25
Move the data-science recipe to contrib/python

happyhuman:
## What

Moves the `data-science` recipe from the retired `python/agents/` root to
`contrib/python/data-science`, and brings it up to the current recipe standard
so it can pass the `contrib/` CI gates it is now subject to for the first time.

The bulk of the diff is a pure `git mv`. The substantive changes are below.

## Why these changes came with the move

`python/agents/` is listed under `frozen_paths` in `.github/policy.yml` and no
longer accepts changes, so the recipe cannot be maintained where it was. Landing
it in `contrib/` newly subjects it to `validate-recipe-structure`,
`python-validate-recipe` and `python-tests`, none of which ran against it before.

- **`manifest.yaml` (new)** — required for every recipe. `ownership.team` and
  `ownership.poc` are deliberately left on the scaffold placeholders, so
  validation keeps failing until a real owner is named.
- **`pyproject.toml`** — removed the recipe-local `[tool.ruff*]` block so the
  root config is the effective one, lowered `requires-python` to `>=3.11` to
  match the interpreter CI locks with, and declared public PyPI as the default
  index. `uv.lock` regenerated to match.
- **`.env.example` / two source files** — declared the environment variables the
  code already read, and sourced the two hardcoded model names from `MODEL_NAME`.
- **`tests/test_runnability.py` (new)** — required by policy. This recipe builds
  its entire agent graph at import time and reaches BigQuery and Vertex AI while
  doing so, so the test stubs those touchpoints; without them the import cannot
  resolve without live credentials.
- **`tests/integration/test_agents.py`** — relocated from `tests/`. It drives the
  agents for real, and at the top level its import aborted pytest collection for
  the whole recipe. `tests/integration/` matches the other contrib recipes and is
  the path CI already ignores.
- **`README.md`** — updated the setup paths the move invalidated.

## Needs a decision before this can go green

**The recipe is 37.9 MB and `contrib/` caps at 10 MB even on the `large: true`
tier**, so `validate structure` still fails on size. Three files account for
almost all of it:

| File | Size |
| --- | --- |
| `data_science/utils/data/train.csv` | 13.6 MB |
| `flights_dataset/flights_dataset_alloydb.sql` | 13.3 MB |
| `flights_dataset/flight_history_table.csv` | 10.2 MB |

Trimming these means deleting or rehosting recipe content, which is an owner
call rather than part of a move, so this PR leaves them in place.

Two smaller items also need a human: the ownership placeholders above, and the
now-dead `Data Science Agent` row in `python/agents/README.md` — that file is
inside a frozen path, so removing the row fails the retired-folder check without
an `allow-legacy-path` label.

## Verification

Run from `contrib/python/data-science` on Python 3.11, with `.env` copied from
`.env.example`, matching what `python-tests.yml` does:

```
uv run pytest --ignore=tests/integration --ignore-glob="**/test_integration.py"
→ 1 passed
```

Also green locally: `ruff format --check`, `ruff check`, `validate readme`,
`validate placement`, `check_env_vars.py`, `check_recipe_pyproject.py`,
`check_frozen_paths.py`, `uv lock --check`. `validate manifest` and
`validate structure` report only the ownership placeholders and the size
failure described above.

The runnability test proves the recipe imports and builds `root_agent`; it does
not exercise any live behaviour.

Resolves b/552504737


happyhuman:
Closed per request.

PR #2582, opened 2026-08-31
fix(deep-search): don't require Cloud credentials in AI Studio mode

### LINKED ISSUE #2574: [BUG]: python/agents/deep-search missing credentials error when run locally
### Name of sample affected

python/agents/deep-search

### Description of issue

The app is looking for credentials even when i want to run locally.

**Fixed by commenting out 5 lines** in `app/__init__.py`

**Working code after change**
```
import os

# import google.auth

# _, project_id = google.auth.default()
# os.environ.setdefault("GOOGLE_CLOUD_PROJECT", project_id or "")
# os.environ.setdefault("GOOGLE_CLOUD_LOCATION", "global")
# os.environ.setdefault("GOOGLE_GENAI_USE_VERTEXAI", "True")

from app.agent import root_agent
```

I suspect `adk api_server` in `make dev-backend` is causing the `import google.auth` to look for vertexai credentials.

### Environment

NA

### Reproduction steps or code

Following the README under section **Alternative: Local development (run from this sample repo)**. 
After running `make dev`, it errors

### Error log

```
raise exceptions.DefaultCredentialsError(_CLOUD_SDK_MISSING_CREDENTIALS)
google.auth.exceptions.DefaultCredentialsError: Your default credentials were not found. To set up Application Default Credentials, see https://cloud.google.com/docs/authentication/external/set-up-adc for more information.
```

rootkiller6788:
Fixes #2574. The deep-search sample called google.auth.default() at import time, so `make dev` died with DefaultCredentialsError on machines without gcloud ADC even when only GOOGLE_API_KEY was set for local AI Studio use (which is the README's default local path).

I moved the project id lookup behind the same GOOGLE_API_KEY check app/config.py already uses, and load .env from the app folder explicitly so the key is seen before the guard runs. Vertex AI mode is unchanged: without GOOGLE_API_KEY it still resolves ADC and fails loudly if there are no credentials.

Also applied the same guard to the old python/agents/deep-search path since the issue is reported there.

Verified locally: importing the app with GOOGLE_API_KEY set no longer crashes, ruff check/format are clean, and the recipe's runnability test passes.


happyhuman (commented):
The samples/recipes are moved to contrib/python and therefore, we no longer accept any contribution to python/agents. Please work with the sample owners to move them to contrib/python

rootkiller6788:
Thanks for the review. I've dropped the change to `python/agents/deep-search/app/__init__.py` — I'd missed that the folder is retired. The branch now only modifies `core/python/deep-search/app/__init__.py`, the maintained copy that the Python recipe CI runs against, so the "Check for work in retired folders" failure should be resolved.

The remaining fix: that file called `google.auth.default()` unconditionally at import time, so the sample crashed with `DefaultCredentialsError` on machines without Cloud ADC even when only `GOOGLE_API_KEY` was set for the local AI Studio flow (the README's default). The project-id lookup now runs only when `GOOGLE_API_KEY` is unset, i.e. Vertex AI mode, matching the existing mode check in `app/config.py`; the `.env` in the `app/` folder is loaded before that guard runs. Vertex AI behavior is unchanged and still fails loudly without credentials.

On contrib/python: the deep-search recipe already has a maintained home at `core/python/deep-search`, so I didn't relocate it as part of this bugfix. Happy to help the sample owners move it if that's the plan.

Could you please take another look?

happyhuman (approved):

PR #2588, opened 2026-09-02
refactor(llm-auditor): migrate to core/kotlin and adk-kotlin 0.9.0

sherryfox:
The only Kotlin recipe still lived under the retired kotlin/agents tree, where CI neither built nor formatted it, and it was pinned to adk-kotlin 0.2.0 — seven minor versions behind. Its web entry point used AdkWebServer, which 0.9.0 deprecated and 1.0 removes. The frozen-path policy exempts deletions and renames precisely so a recipe can be moved out, so this moves it.

It lands in core/kotlin rather than contrib/kotlin. contrib is the community tier — its recipes are owned by individuals and outside organisations — while core holds recipes curated by first-party teams, and this is one. core requires an AGENTS.md, which this adds.

WebMain.kt now builds an AdkDevServer from an AdkServerConfig. AdkServerConfig .inMemory() supplies the agent loader and the in-memory session and artifact services the old constructor took one by one, so three imports go. The old code passed ApiServerSpanExporter() explicitly; the new constructor defaults it to the same thing, so the Dev UI trace view still works. The new classes bind loopback where AdkWebServer pinned 0.0.0.0, so the README says so and shows the host override.

Two follow-on corrections the move forces:

  - kotlin-format.yml cited kotlin/README.md as documenting the JDK 17 minimum and said no core or contrib Kotlin recipe existed yet. Both stop being true here; the dependabot comment naming kotlin/agents/llm-auditor/build.gradle .kts as a live path likewise. Dependabot's config globs "**/*", so only the prose was stale.
  - docs/recipe-handbook/languages/kotlin.md opened with "No Kotlin recipes have landed yet and no CI is in place". It now describes the tooling that does apply.

kotlin/agents is now empty, so kotlin/README.md points at the two live roots instead. Its "browse agents/" instructions pointed at nothing, and it carried a dead ../android/ link.

Verified: the recipe compiles against published 0.9.0 via `gradle test`, the exact command kotlin-tests.yml runs; ktlint 1.8.0 (the pinned version, checksum-matched) is clean, which it was not before — the four pre-existing files had ten findings that had never been checked, since kotlin/agents was excluded from that workflow; `uv run validate` passes all four checks, and removing AGENTS.md was confirmed to fail structural validation, so the core requirement is genuinely met. The recipe has no tests, so `gradle test` proves compilation, not behaviour.


.github/workflows/kotlin-format.yml
happyhuman, line 1: Was there a reason this file was updated?

.github/dependabot.yml
happyhuman, line 1: Was there a reason this file was updated?

docs/recipe-handbook/languages/kotlin.md
happyhuman, line 1: Was there a reason this file was updated?

docs/recipe-handbook/languages/kotlin.md
```diff
@@ -1,20 +1,21 @@
...
+Early days. `core/kotlin/llm-auditor` is the first recipe here,
+and there are no Kotlin authoring skills yet.
 
-**If you want to contribute one:** open a GitHub issue at
-[github.com/google/adk-samples/issues](https://github.com/google/adk-samples/issues)
-first so we can align on build tool (Gradle / Maven), test
-runner, and JVM target before you invest the work. Once accepted,
-this page will mirror the shape of the [Python page](./python.md).
+**Tooling that applies today:** structural checks (folder name, size
+limits, `manifest.yaml`); `kotlin-format.yml`, which runs ktlint over
```
happyhuman: No need to mention internal workflows in the docs.

happyhuman (commented):
Just left a few comments.

.github/workflows/kotlin-format.yml
sherryfox: Reverted

.github/dependabot.yml
sherryfox: Reverted

docs/recipe-handbook/languages/kotlin.md
sherryfox: One sentence stops being true when this merges: the page opens "Not currently active. No Kotlin recipes have landed yet", and kotlin/README.md — which this PR rewrites — links contributors straight here. Leaving it would have the handbook telling people Kotlin has no recipes on the day the first one lands.

happyhuman (approved):

PR #2621, opened 2026-09-10
feat(ci): publish recipe container images to Artifact Registry

haomengchao:
make-python-recipe-deployable writes the serving files a recipe needs —
Dockerfile, fast_api_app.py, and the app_utils/ adapters implementing the
Agent Engine container contract — and then stops, on the stated assumption
that "image builds happen later via Cloud Build → Artifact Registry".
Nothing did that later step, so a recipe could satisfy every deployability
check and still have no image anyone could pull.

That gap has a consumer waiting on it. Agent Engine can deploy an agent
directly from a prebuilt image via container_spec.image_uri, which skips the
multi-minute per-deployment source build that App Design Center pays on every
deploy today. The saving is only available to recipes that have a published
image.

recipe_images_matrix.py decides what gets built. Three constraints shape it,
each of which produces a wrong answer if dropped:

  * Live roots only. SCAN_ROOTS is imported from recipe_manifests rather than
    restated, because the retired roots in policy.yml frozen_paths still
    contain Dockerfiles and would otherwise publish images from code that is
    closed to fixes.
  * Recipe-root Dockerfiles only. multiformat-hybrid-rag ships three; two are
    data-ingestion sub-services with their own lifecycles, and publishing
    them under a recipe image name would misrepresent what the image is.
  * A manifest.yaml must sit beside the Dockerfile. Without that test,
    skills/retail/virtual-tryon/assets/export-template is mistaken for a
    recipe and derives "retail" as its language.

Images are named <language>/<recipe>, dropping the root segment so that
promoting a recipe from contrib/ to core/ does not rename the image out from
under anyone who pinned it. Language is retained because core/kotlin/
llm-auditor and contrib/python/llm-auditor coexist.

The workflow rebuilds only what a push touched, and everything weekly — the
Dockerfiles pin python:3.12-slim and resolve dependencies at build time, so
an image that built cleanly last month can fail today against a patched base
layer or a yanked release, which no push-triggered run would ever notice.

Authentication is Workload Identity Federation against the pool already
configured for this repository; no key material is introduced. The target
repository grants allUsers artifactregistry.reader, which is what lets Agent
Engine pull an image into a customer tenant without a per-consumer IAM grant.
The images hold only Apache-2.0 sample code already published here.

### DIFF
--- a/.github/scripts/recipe_images_matrix.py
+++ b/.github/scripts/recipe_images_matrix.py
@@ -70,11 +70,13 @@ from recipe_manifests import REPO_ROOT, SCAN_ROOTS, SKIP_DIRS
 REGISTRY = "us-west1-docker.pkg.dev/adk-samples-repo-gcp-support/adk-recipes-registry"
 
 
-def _language_and_name(recipe: Path) -> tuple[str, str]:
-    """Split a recipe path into (language, name).
+def _category_and_name(recipe: Path) -> tuple[str, str]:
+    """Split a recipe path into (category, name).
 
-    Recipes live at <root>/<language>/<name>. Anything shallower than that has
-    no language segment to report, so the root stands in for it.
+    Recipes live at <root>/<category>/<name>. The segment is a language under
+    core/ and contrib/ (python, kotlin) but a vertical under skills/ (retail),
+    so it is named for what it is positionally rather than for what it usually
+    holds. Anything shallower has no such segment, and the root stands in.
     """
     parts = recipe.parts
     if len(parts) >= 3:
@@ -83,13 +85,19 @@ def _language_and_name(recipe: Path) -> tuple[str, str]:
 
 
 def discover(repo_root: Path) -> list[dict[str, str]]:
-    """Every live recipe with a Dockerfile at its root, sorted by path."""
+    """Every live recipe with a Dockerfile at its root, shallowest first."""
     found: list[dict[str, str]] = []
     for root in SCAN_ROOTS:
         base = repo_root / root
         if not base.is_dir():
             continue
-        for dockerfile in sorted(base.rglob("Dockerfile")):
+        # Depth first, not lexicographic. The nested-Dockerfile test below
+        # only works if a recipe is recorded before anything beneath it, and
+        # sorting by name does not guarantee that: 'foo/Bar/Dockerfile' sorts
+        # ahead of 'foo/Dockerfile' because 'B' < 'D'. Depth does guarantee
+        # it, and the secondary sort keeps the output stable.
+        for dockerfile in sorted(base.rglob("Dockerfile"),
+                                 key=lambda p: (len(p.parts), p)):
             recipe = dockerfile.parent
             rel = recipe.relative_to(repo_root)
             if any(part in SKIP_DIRS for part in rel.parts):
@@ -98,37 +106,42 @@ def discover(repo_root: Path) -> list[dict[str, str]]:
                 continue
             # A Dockerfile nested under another recipe belongs to a
             # sub-service, not to a recipe of its own.
-            if any(entry["path"] != str(rel) and str(rel).startswith(entry["path"] + "/")
-                   for entry in found):
+            if any(str(rel).startswith(entry["path"] + "/") for entry in found):
                 continue
-            language, name = _language_and_name(rel)
+            category, name = _category_and_name(rel)
             found.append(
                 {
                     "path": str(rel),
                     "name": name,
-                    "language": language,
-                    "image": f"{REGISTRY}/{language}/{name}",
+                    "category": category,
+                    "image": f"{REGISTRY}/{category}/{name}",
                 }
             )
     return found
 
 
-def _changed_paths(ref: str) -> set[str]:
-    """Files touched between `ref` and HEAD, as repo-relative paths."""
+def _changed_paths(ref: str, repo_root: Path) -> set[str] | None:
+    """Files touched between `ref` and HEAD, or None if the diff failed.
+
+    None and the empty set mean different things and the caller acts on the
+    difference: an empty set is a successful diff that found nothing, so
+    nothing needs rebuilding, while None means the question could not be
+    answered and every recipe should be rebuilt rather than silently skipped.
+    """
     out = subprocess.run(
         ["git", "diff", "--name-only", f"{ref}...HEAD"],
-        cwd=REPO_ROOT,
+        cwd=repo_root,
         capture_output=True,
         text=True,
         check=False,
     )
     if out.returncode != 0:
         # A missing or unrelated ref is not worth failing the build over:
-        # falling back to "everything changed" errs toward rebuilding, which
-        # is wasteful but never publishes a stale image.
+        # falling back to "everything changed" is wasteful but never leaves a
+        # recipe published from stale source.
         print(f"warning: git diff against {ref} failed, treating all recipes as changed",
               file=sys.stderr)
-        return set()
+        return None
     return {line.strip() for line in out.stdout.splitlines() if line.strip()}
 
 
@@ -156,8 +169,11 @@ def main() -> int:
             return 1
 
     if args.changed_from:
-        changed = _changed_paths(args.changed_from)
-        if changed:
+        changed = _changed_paths(args.changed_from, REPO_ROOT)
+        # None is a failed diff: rebuild everything rather than skip silently.
+        # An empty set is a successful diff that found nothing, which
+        # correctly narrows the matrix to nothing.
+        if changed is not None:
             recipes = [r for r in recipes
                        if any(f.startswith(r["path"] + "/") for f in changed)]
 
--- a/.github/scripts/recipe_images_matrix.py
+++ b/.github/scripts/recipe_images_matrix.py
@@ -16,7 +16,7 @@ half of that later step: it turns the tree into a build matrix.
 Recipe-root Dockerfiles only
 ----------------------------
 A recipe can contain more than one Dockerfile. contrib/python/multiformat-
-hybrid-rag ships three — one for the recipe, two for data-ingestion
+hybrid-rag ships four — one for the recipe, three for data-ingestion
 sub-services it stands up. Only the file at the recipe root serves the agent
 and satisfies the Agent Engine container contract; the others are backing
 infrastructure with their own lifecycles. Publishing those under a recipe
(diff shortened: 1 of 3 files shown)


.github/scripts/recipe_images_matrix.py
```diff
@@ -0,0 +1,169 @@
...
+    )
+    args = parser.parse_args()
+
+    recipes = discover(REPO_ROOT)
+
+    if args.recipe:
+        wanted = args.recipe.rstrip("/")
+        recipes = [r for r in recipes if r["path"] == wanted]
+        if not recipes:
+            print(f"error: no recipe with a root Dockerfile at {wanted!r}", file=sys.stderr)
```
happyhuman: this file isn't ruff-formatted

.github/scripts/recipe_images_matrix.py
```diff
@@ -0,0 +1,169 @@
...
+    Recipes live at <root>/<language>/<name>. Anything shallower than that has
+    no language segment to report, so the root stands in for it.
+    """
+    parts = recipe.parts
+    if len(parts) >= 3:
+        return parts[1], parts[-1]
+    return parts[0], parts[-1]
+
+
+def discover(repo_root: Path) -> list[dict[str, str]]:
```
happyhuman, line 100: Every other script in `.github/scripts/` has a `test_*.py` beside it — `recipe_canary_matrix.py` has `test_recipe_canary_matrix.py`. Is one coming for this module?

.github/scripts/recipe_images_matrix.py
```diff
@@ -0,0 +1,169 @@
...
+---------------------------------
+`.agents/skills/make-python-recipe-deployable` writes the serving files a
+recipe needs — Dockerfile, fast_api_app.py, app_utils/ — and then stops. Its
+own words: "Image builds happen later via Cloud Build → Artifact Registry;
+this skill's job ends when the files are correct." This module is the front
+half of that later step: it turns the tree into a build matrix.
+
+Recipe-root Dockerfiles only
+----------------------------
+A recipe can contain more than one Dockerfile. contrib/python/multiformat-
```
happyhuman, line 18: four Dockerfiles under that recipe, not three

.github/workflows/recipe-images.yml
```diff
@@ -0,0 +1,215 @@
...
+# Engine pull the image with its own service agent, in a project we have
+# never heard of. Public read is what removes the otherwise unbounded set of
+# per-consumer IAM grants. These images contain only Apache-2.0 sample code
+# already published in this repository — no secrets, no customer data.
+#
+# Three triggers, two different questions
+# ---------------------------------------
+# `push` asks "did this change break its own image?" and rebuilds only the
+# recipes the push touched. `schedule` asks a question no push can: the
+# Dockerfiles pin python:3.12-slim and resolve dependencies at build time, so
```
happyhuman: I see `python:3.11-slim` in `contrib/python/multiformat-hybrid-rag/Dockerfile` and `uv:python3.11-bookworm-slim` in `core/python/ambient-expense-agent/Dockerfile`. Is 3.12 the right version to name here?

.github/workflows/recipe-images.yml
```diff
@@ -0,0 +1,215 @@
...
+        with:
+          context: ${{ matrix.path }}
+          file: ${{ matrix.path }}/Dockerfile
+          # `latest` for humans and for docs that cannot chase a digest;
+          # the commit tag so a deployed image can be traced to the source
+          # that produced it. A dry run builds both and pushes neither.
+          push: ${{ !inputs.dry_run }}
+          tags: |
+            ${{ matrix.image }}:latest
+            ${{ matrix.image }}:${{ github.sha }}
```
happyhuman: The weekly run tags with `github.sha` as well, so an unchanged commit gets its `:sha` tag rewritten every Tuesday. Is that intentional?

.github/scripts/recipe_images_matrix.py
```diff
@@ -0,0 +1,169 @@
...
+        recipes = [r for r in recipes if r["path"] == wanted]
+        if not recipes:
+            print(f"error: no recipe with a root Dockerfile at {wanted!r}", file=sys.stderr)
+            return 1
+
+    if args.changed_from:
+        changed = _changed_paths(args.changed_from)
+        if changed:
+            recipes = [r for r in recipes
+                       if any(f.startswith(r["path"] + "/") for f in changed)]
```
happyhuman: A push touching only this script or `recipe-images.yml` matches no recipe prefix here, though both are in the workflow's `paths` filter. Does the matrix just come back empty then?

happyhuman (commented):
Left a few comments. PTAL.

.github/scripts/recipe_images_matrix.py
haomengchao: Fixed — four. One recipe plus three under `data_ingestion_pipeline/` (the pipeline itself, `chunk_index_service`, `preprocess_service`).

.github/scripts/recipe_images_matrix.py
haomengchao: Fixed, both files are ruff-formatted now.

.github/workflows/recipe-images.yml
haomengchao: No, 3.12 was wrong. Across the five recipes in scope: `python:3.11-slim` (multiformat-hybrid-rag, cross-session-memory), `python:3.12-slim` (long-horizon-harness), `python:3.13-slim` (genmedia-for-commerce), `ghcr.io/astral-sh/uv:python3.11-bookworm-slim` (ambient-expense-agent). I'd generalised from the template Dockerfile in `make-python-recipe-deployable` rather than from the recipes. Rewritten to state the actual spread, which argues for the weekly rebuild better than a single wrong version did.

.github/workflows/recipe-images.yml
haomengchao: Not intentional, and it's a real defect — thanks. The weekly run exists to produce a *different* image from an unchanged commit (fresh base layers, re-resolved dependencies), so `:<sha>` was being handed content that commit never described, and anything pinned to it moved silently every Tuesday.

The pinnable tag is now `<date>-<short-sha>`, e.g. `20260910-1b48f2ccb56c`. Immutable, still traceable to source, and the weekly rebuild gets its own. Computed once in the matrix job rather than per leg, so a matrix straddling midnight UTC doesn't split across two dates. `:latest` still moves.

.github/scripts/recipe_images_matrix.py
haomengchao: It did, and the build job's `if: count != '0'` then skipped it entirely — so a change to the matrix logic could merge having never built anything, which is exactly when you'd want it to. Fixed: both files are now in `SELF_PATHS`, and a diff touching either skips the incremental filter and rebuilds the full set. Pinned in `test_self_paths_cover_both_files_that_define_the_matrix`.

.github/scripts/recipe_images_matrix.py
haomengchao: It is now — `tests/test_recipe_images_matrix.py`, 10 cases. Aimed at the admission rules rather than the happy path, since both directions of a wrong answer are silent here: publishing an image from a build asset, or leaving a recipe with none. Covers the manifest requirement, retired-root exclusion, nested sub-services, image-name stability across a contrib/→core/ promotion, and the ordering regression from the commit before it.

.github/scripts/recipe_images_matrix.py
```diff
@@ -0,0 +1,169 @@
...
+import subprocess
+import sys
+from pathlib import Path
+
+from recipe_manifests import REPO_ROOT, SCAN_ROOTS, SKIP_DIRS
+
+# Every image lands here. The repository is public-read (allUsers has
+# roles/artifactregistry.reader), which is what lets Agent Engine pull an
+# image into a customer tenant without a per-consumer IAM grant.
+REGISTRY = "us-west1-docker.pkg.dev/adk-samples-repo-gcp-support/adk-recipes-registry"
```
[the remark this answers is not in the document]
haomengchao: Leaving these in the source deliberately. The project and registry aren't credentials — they're the published destination of this workflow, and the thing a reviewer most needs to see when judging whether the change is safe. Moving them into repository variables hides who changed the publish target and when, behind a settings page with no review trail, in exchange for no security benefit. The workflow already collects them in one `env:` block at the top.

.github/scripts/recipe_images_matrix.py
```diff
@@ -0,0 +1,169 @@
...
+            # sub-service, not to a recipe of its own.
+            if any(entry["path"] != str(rel) and str(rel).startswith(entry["path"] + "/")
+                   for entry in found):
+                continue
+            language, name = _language_and_name(rel)
+            found.append(
+                {
+                    "path": str(rel),
+                    "name": name,
+                    "language": language,
```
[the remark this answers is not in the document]
haomengchao: Kept, and renamed to `category` per the sibling comment. It's what `image` is built from, and it makes the matrix output legible when debugging a run — the alternative is re-deriving it from `path` by eye. The matrix contract is already three keys wide; a fourth that explains the third seems worth it.

.github/scripts/tests/test_recipe_images_matrix.py
```diff
@@ -0,0 +1,151 @@
...
+
+    A failed diff must rebuild everything; a clean diff that found nothing
+    must rebuild nothing. Collapsing both to an empty set makes the second
+    case behave like the first.
+    """
+    assert m._changed_paths("no-such-ref", tmp_path) is None
+
+
+def test_self_paths_cover_both_files_that_define_the_matrix(
+    tmp_path: Path,
```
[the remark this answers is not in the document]
haomengchao: Fixed — the assertion reads a module constant and touches no filesystem, so the fixture was requested and never used.

.github/scripts/recipe_images_matrix.py
```diff
@@ -0,0 +1,207 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
haomengchao: Not adding one: no script in `.github/scripts/` carries it. Checked `recipe_manifests.py`, `recipe_canary_matrix.py`, `check_env_vars.py` and `load_policy.py` — none have the header. The `tests/` files do, and `tests/test_recipe_images_matrix.py` in this PR has it. This matches the convention on both sides.

.github/workflows/recipe-images.yml
```diff
@@ -0,0 +1,215 @@
...
+
+concurrency:
+  # Serialized per ref rather than cancelled: a half-finished matrix leaves
+  # some recipes published at the new commit and some at the old one, which
+  # is a worse state to debug than a queued run.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
+env:
+  WORKLOAD_IDENTITY_PROVIDER: projects/148460639759/locations/global/workloadIdentityPools/adk-samples-github-pool/providers/adk-samples-github-provider
```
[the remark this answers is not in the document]
haomengchao: Same answer as the earlier thread on this: the project number and registry path aren't credentials, they're the published destination of the workflow, and keeping them in the diff is what lets a reviewer see the publish target change. Moving them to repository variables trades that review trail for no security benefit. They're collected in one `env:` block at the top of the workflow.

happyhuman (approved):

PR #2622, opened 2026-09-11
Upgrade region and dependencies for deep-search recipe

happyhuman:
## What

Updates the `deep-search` recipe under `core/python/deep-search/` to modernize its dependencies and configuration. Updates the default region in documentation to `us-east1`, raises the minimum `google-adk` requirement to `>=1.31.0` (resolving to `2.9.0`), refreshes `uv.lock`, removes hardcoded model fallbacks in `app/config.py`, and updates recipe authors in `pyproject.toml`.

## Why

The recipe was pinned to older ADK versions (`>=1.8.0`) and had stale lockfile dependencies. Additionally, GCP location guidance needed to reflect `us-east1`, recipe metadata was missing ownership authors from the manifest, and `app/config.py` contained hardcoded model defaults in `os.getenv()` reads instead of relying on `.env.example`.

## How it works

- `core/python/deep-search/pyproject.toml`: Added `authors` section with the Google POC from `manifest.yaml` (`Elia Secchi <eliasecchi@google.com>`), and updated `google-adk>=1.31.0`.
- `core/python/deep-search/app/config.py`: Replaced hardcoded fallback strings in `os.getenv("MODEL_NAME")` with bare reads, delegating the default configuration to `.env.example`.
- `core/python/deep-search/README.md`: Updated Vertex AI setup snippet `GOOGLE_CLOUD_LOCATION` from `us-central1` to `us-east1`.
- `core/python/deep-search/uv.lock`: Upgraded via `uv lock --upgrade`.
- `core/python/deep-search/tests/test_runnability.py`: Applied formatting to match ruff rules.

## Testing

Verified with pytest and full lint suite (`make lint`):

```
uv sync --dev --extra lint
uv run codespell
uv run ruff check . --diff
uv run ruff format . --check --diff
6 files already formatted
uv run mypy .
Success: no issues found in 4 source files
uv run pytest tests/test_runnability.py
======================== 1 passed, 5 warnings in 1.15s =========================
```

Note: Verification tests import runnability and static analysis; live end-to-end model execution and Vertex deployment are not run locally.

## Notes for the reviewer

Left the `LoopAgent`/`SequentialAgent` structure intact in `app/agent.py`. While ADK v2 deprecates these in favor of `Workflow`, ADK notes `Workflow` cannot yet be used as an `LlmAgent` sub-agent, so the existing agent architecture is preserved.

### DIFF
--- a/core/python/deep-search/app/config.py
+++ b/core/python/deep-search/app/config.py
@@ -52,8 +52,8 @@ class ResearchConfiguration:
         max_search_iterations (int): Maximum search iterations allowed.
     """
 
-    critic_model: str = os.getenv("MODEL_NAME")
-    worker_model: str = os.getenv("MODEL_NAME")
+    critic_model: str = os.getenv("MODEL_NAME", "gemini-3.5-flash")
+    worker_model: str = os.getenv("MODEL_NAME", "gemini-3.5-flash")
     max_search_iterations: int = 5
 
 
--- a/core/python/deep-search/pyproject.toml
+++ b/core/python/deep-search/pyproject.toml
@@ -1,10 +1,13 @@
 [project]
 name = "deep-search"
 version = "0.1.0"
+authors = [
+    { name = "Elia Secchi", email = "eliasecchi@google.com" },
+]
 requires-python = ">=3.11,<3.13"
 
 dependencies = [
-    "google-adk>=1.8.0",
+    "google-adk>=1.31.0",
     "python-dotenv>=1.0.0",
 ]
 
--- a/core/python/deep-search/README.md
+++ b/core/python/deep-search/README.md
@@ -107,7 +107,7 @@ If you've already cloned the repository and want to use Vertex AI instead of AI 
 ```bash
 echo "GOOGLE_GENAI_USE_VERTEXAI=TRUE" >> app/.env
 echo "GOOGLE_CLOUD_PROJECT=YOUR_PROJECT_ID" >> app/.env
-echo "GOOGLE_CLOUD_LOCATION=us-central1" >> app/.env
+echo "GOOGLE_CLOUD_LOCATION=us-east1" >> app/.env
 ```
 
 Make sure you're authenticated with Google Cloud:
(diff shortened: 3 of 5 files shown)


core/python/deep-search/app/config.py
```diff
@@ -52,8 +52,8 @@ class ResearchConfiguration:
         max_search_iterations (int): Maximum search iterations allowed.
     """
 
-    critic_model: str = os.getenv("MODEL_NAME")
-    worker_model: str = os.getenv("MODEL_NAME")
+    critic_model: str = os.getenv("MODEL_NAME", "gemini-3.5-flash")
```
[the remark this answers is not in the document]
happyhuman: Removed the hardcoded fallback defaults from `os.getenv("MODEL_NAME")` in `app/config.py` in favor of `.env.example`.

core/python/deep-search/pyproject.toml
```diff
@@ -1,10 +1,13 @@
 [project]
 name = "deep-search"
 version = "0.1.0"
+authors = [
```
happyhuman: authors should be in manifest.yaml file and not in here. Make sure Elia's github ID is properly set as poc in manifest.yaml and remove this from pyproject.toml

core/python/deep-search/pyproject.toml
```diff
@@ -1,10 +1,13 @@
...
+    { name = "Elia Secchi", email = "eliasecchi@google.com" },
+]
 requires-python = ">=3.11,<3.13"
 
 dependencies = [
-    "google-adk>=1.8.0",
+    "google-adk>=1.31.0",
```
happyhuman: Can we safely update this to version 2?

core/python/deep-search/pyproject.toml
happyhuman: Removed the `authors` block from `pyproject.toml`. `manifest.yaml` already has Elia set correctly as `poc: "eliasecchig"`, so no change was needed there.

core/python/deep-search/pyproject.toml
happyhuman: Yes — it was already on v2 in practice. The lockfile resolved `>=1.31.0` to `google-adk` 2.9.0, so the declared floor claimed v1 support that nothing had been tested against. Tightened it to `google-adk>=2.9.0,<3.0.0`, matching `rag-agent-search` / `rag-vector-search` / `ambient-expense-agent`. Relocking changed no dependency versions, only the specifier line.

Verified on 2.9.0: `mypy` clean across 4 source files and `pytest tests/test_runnability.py` passes. The only warnings are the known `LoopAgent` / `SequentialAgent` deprecations, which stay as-is since `Workflow` cannot yet be an `LlmAgent` sub-agent.

PR #2640, opened 2026-09-17
fix(ci): give the AI review job a genuinely read-only token

happyhuman:
## The problem

The `review` job in `_ai-pr-review-core.yml` holds the Google Cloud credential
and runs a model over a diff an anonymous fork author wrote. It declares:

```yaml
permissions: {contents: read, id-token: write, pull-requests: read}
```

It was nonetheless holding a **write-capable** token on every run.

The core's DESIGN note already documented this, as a caveat that applies "when
`app_id` is empty". That turns out to describe the only configuration this repo
has ever run in — **`vars.APP_ID` is not set**, so `Generate GitHub App Token`
is `skipped` on every run and the fallback is the only path that executes.
Confirmed on a live run:

```
$ gh api .../actions/runs/<latest ai-pr-review-security>/jobs
Generate GitHub App Token (read-only): skipped
```

The fallback was `secrets.GITHUB_TOKEN_SECRET`, passed in by the calling lane,
whose `permissions:` grants `issues: write` and `pull-requests: write`. **A
secret passed *into* a job is not re-scoped by that job's own `permissions:`**,
so the declared `pull-requests: read` bought nothing.

## The fix

Switch the fallback to `github.token`, which *is* re-scoped per job.

| job | declared permissions | token scope after |
|---|---|---|
| `review` | `contents: read, id-token: write, pull-requests: read` | **read-only** |
| `post` | `issues: write, pull-requests: write` | write |
| `notify` | `issues: write, pull-requests: write` | write |

One step (`Filter the diff and size the review`) already used `github.token`,
which is what proved the mechanism works in this context before moving the
other nine.

With no consumer left, `GITHUB_TOKEN_SECRET` is removed from the core's
`workflow_call.secrets` and from all four lanes — so the write-capable token is
no longer passed across the job boundary at all.

The App-token path is untouched and still takes precedence wherever `APP_ID`
is configured.

## What this does *not* change

The lanes keep their top-level `issues: write, pull-requests: write`. For a
reusable workflow the caller's grant is the *ceiling* for the callee's jobs, so
`post` and `notify` still need it. The bug was specifically the secret
bypassing per-job scoping, and that is what is fixed.

Setting `APP_ID` is still worth doing — it would restore the App-token path and
give `review` a token that is read-only by construction rather than by scope —
but it needs an App ID and private key I don't have.

## Verification

- `uv run pytest -q` → **1991 passed**
- `actionlint` → 0 non-shellcheck findings (SC2086/SC2001 unchanged from baseline)
- Caller/callee secret contract re-resolved in both directions: core declares
  `[APP_PRIVATE_KEY]`, all four lanes pass exactly `[APP_PRIVATE_KEY]`
- Programmatic walk of the job graph confirms no step in `review` references a
  caller-minted token, and `post`/`notify` retain write scope

---


happyhuman:

PR #2641, opened 2026-09-17
fix(ci): stop the Docker gate passing a container that serves nothing

happyhuman:
## The bug

`validate_recipe_docker` accepted **HTTP 404** as proof that a recipe's
container is accessible:

```python
if status is not None and (200 <= status < 400 or status in (401, 403, 404)):
```

A 404 says only that *some process is speaking HTTP on the port*. It does not
say the ADK app mounted. An image whose web server boots but whose app never
starts answers 404 on **every** probe path — and this gate called that a PASS.
That is precisely the false green the check exists to prevent.

## The fix

404 is excluded. `401`/`403` are still accepted — refusing a request requires
the route to exist and the request to have been routed to it, which is the
signal we want. 2xx/3xx likewise.

**This tightens nothing that currently passes.** All six recipes with a root
Dockerfile answer `/list-apps` with HTTP 200 on the green `main` run
([35176298541](https://github.com/google/adk-samples/actions/runs/35176298541)),
so none was relying on the leniency.

Also realigns the probe timeout with `.github/policy.yml`, which declares
`deployability.verification` the source of truth and says
`ready_timeout_seconds: 120` where this module had drifted to `90`.

## ⚠️ One divergence I did NOT resolve — needs a maintainer decision

`policy.yml` defines `deployability.verification.run_allowlist`:

> Recipes whose containers are safe to RUN, not merely build. Some recipes
> create real GCP resources at import — for example
> `core/python/cross-session-memory` calls `client.agent_engines.create()` at
> module scope. [...] anything not listed here is BUILT ONLY.

**CI honours no such list.** It runs every recipe with a root Dockerfile, and
there is *zero overlap* between the two sets:

| | |
|---|---|
| policy allows running | `contrib/python/financial-advisor`, `core/python/rag-agent-search`, `core/python/rag-vector-search` |
| CI actually ran | `brand-search-optimization`, `multiformat-hybrid-rag`, `ambient-expense-agent`, **`cross-session-memory`**, `genmedia-for-commerce`, `long-horizon-harness` |

Including the one recipe the policy names as its example of the hazard.

It is harmless *today* only because the job holds no real credentials — the
container gets a placeholder project and no ADC, which is why the unguarded
telemetry bug surfaced as a crash rather than as a provisioned resource. That
is safety by accident, not by design. Wire real credentials into that job and
the policy's worst case becomes live.

Resolving it is a genuine choice between two costs:

- **Honour the allowlist** → all six recipes become build-only, losing the run
  verification that just caught the telemetry bug on two of them.
- **Rescope the allowlist** to credentialed contexts only, and say so in policy.

That is your call, not a bug-fix's, so it is documented in a comment in the
source and left alone.

## Verification

- 4 regression tests added, and I checked they earn their place: with the old
  behaviour restored, `test_all_paths_404_is_not_accessible` and
  `test_probe_timeout_default_matches_policy` **both fail**; with the fix both
  pass.
- `test_probe_timeout_default_matches_policy` reads `policy.yml` directly, so
  the next drift is a test failure rather than a silent difference.
- `uv run pytest -q` → **1995 passed** (up from 1991)
- ruff clean

---


happyhuman:

PR #2642, opened 2026-09-17
fix(ci): stop the reviewer discarding findings it could post

happyhuman:
Two silent losses in `post_review_comments.py`. Both blamed the model for
something the parser could not read.

## 1. Truncated output was a hard CI fault

`FENCED_BLOCK` needs a closing fence to match, and the bare-`[` fallback does
not fire when the response opens with the fence. So a response cut off
mid-array — the ordinary shape of an output-limit hit — raised
`response contained no JSON findings array` → **exit 2** → a red check and a
failure comment on the contributor's PR.

The repair and salvage paths downstream already cope with a block that stops
mid-object. Nothing was reaching them.

Demonstrated, identical content both ways:

```
truncated (no closing fence)  ->  ReviewerOutputError: response contained no JSON findings array
+ closing fence               ->  salvaged 1 finding(s) from it
```

An unterminated block is now handed to the same path, so truncation degrades to
"post what parsed". A response with genuinely no array still raises.

## 2. Git-quoted paths dropped every finding in the file

With `core.quotePath` on — the default — git quotes and C-escapes any path
containing a space, a quote or a non-ASCII byte:

```
+++ "b/my recipe/agent.py"
```

Two things fail together: the quotes stay in the key, and the `b/` prefix is
**inside** them, so the prefix strip never fires. The diff was keyed under
`"b/my recipe/agent.py"` while the API reports `my recipe/agent.py`, and every
finding in that file was discarded as unanchorable.

```
before:  keyed under ['"b/my recipe/agent.py"']   match? False
after:   keyed under ['my recipe/agent.py']       match? True
```

`prepare_review_diff.py` already fixed exactly this on its side of the pipeline,
with tests. I **import** those helpers rather than re-implementing them —
duplicating this logic is how the two drifted apart and left this module unable
to read a path its sibling handles correctly.

Order matters and is now asserted: unquote first, strip the side prefix second.
Verified across four header shapes:

| header | key | note |
|---|---|---|
| `+++ "b/my recipe/agent.py"` | `my recipe/agent.py` | space |
| `+++ "b/caf\303\251.py"` | `café.py` | non-ASCII |
| `+++ b/app/agent.py` | `app/agent.py` | plain |
| `+++ b/b/thing.py` | `b/thing.py` | real top-level `b/` dir — must keep its prefix |

## Verification

- 6 regression tests added. With the fixes reverted **3 fail**; with them, all pass.
- `uv run pytest -q` → **1997 passed** (1991 on main + 6 new)
- ruff clean

---


happyhuman:

PR #2643, opened 2026-09-17
fix(ci): stop the branch sweep deleting a stack's base an hour after its PR closes

happyhuman:
## The hazard

`open_pr_protected_refs` protects a base ref only while its PR is **OPEN**. The
Monday ordering then opens a one-hour window:

```
05:00  stale-sweep.yml         closes stale PRs
06:00  stale-branch-sweep.yml  deletes stale branches
```

At 06:00 a branch that was only ever a stack's **base** has lost that
protection, and nothing else credits it — `prs_for_branch` filters on
`head_ref`, so the just-closed PR counts for the *child* branch, not the base.

It therefore fell through to the **orphan clock**, measured from its last
commit. That commit is necessarily old: being idle is *why* the PR went stale.
So a stack idle past `orphan_after_days` could have its base deleted within the
hour of losing protection.

That also falsified the notice `stale-sweep.yml` posts an hour earlier:

> This is housekeeping, not a verdict on the change. If you want to pick it up
> again, **reopen the pull request — your commits are unaffected**

Not true once the base of the stack is gone. This is the same argument
`stale-sweep.yml` already makes for not deleting head refs — it had simply
never been applied to base refs.

## The fix

A branch that is the base of a recently-closed PR now takes the **closed-PR
clock**, which runs from `closedAt` — exactly the "it may still be revived"
window that message promises.

Cross-repository PRs count here, unlike for head refs, because a base ref
always names a branch in *this* repository. That is the same asymmetry
`open_pr_protected_refs` already relies on.

**This defers deletion, it does not prevent it.** Once `closed_pr_after_days`
has passed the base is deleted as before, and a branch no PR ever referenced is
untouched. Both are asserted.

## Verification

- 4 tests added. With the fix reverted, **3 fail** — including the fork case and
  the bounded-grace case, so this is not just testing the happy path.
- `uv run pytest -q` → **1995 passed**
- ruff clean

---


happyhuman:

PR #2644, opened 2026-09-17
fix(ci): add the missing credential scan to the issue-triage core

happyhuman:
## The gap

Three reusable cores hold a Google Cloud credential and run a model over text
this repository does not control. Only one had **no credential scan**:

| core | empty tool allowlist | response not printed | credential scan |
|---|:--:|:--:|:--:|
| `_ai-pr-review-core.yml` | ✅ | ✅ | ✅ |
| `_ai-issue-response-core.yml` | ✅ | ✅ | ✅ |
| `_ai-issue-triage-core.yml` | ✅ | ✅ | ❌ |

It was also absent from the two parametrized hardening tests that assert the
other two cores have it — so nothing flagged the omission.

This core runs a model over **issue bodies, which anyone can write**, on an
hourly-ish schedule, holding the same ADC as the other two.

## The fix

Matches the issue-response core: needles built from
`ACTIONS_ID_TOKEN_REQUEST_TOKEN` and the secret-bearing fields of
`$GOOGLE_APPLICATION_CREDENTIALS`, built **before** the model runs so an early
exit cannot skip them, and refusing to emit anything if they appear.

Public config fields stay out of the needle list on purpose — `.type`,
`.token_url`, `.audience`, `.quota_project_id` appear legitimately in this
repo's own Google Cloud recipes, and scanning for them would fail honest runs
and cry security incident.

### One deliberate improvement over the other two

This scan also covers **`agy_stderr.log`**, not just `agy_result.json`.

This core `cat`s that file straight into a world-readable workflow log a few
lines further down. An unscanned stderr would have been the easiest way out of
the runner — and it is the same gap the PR-review core still has, where the
scan runs ~40 lines *after* the stderr is printed. Worth fixing there too, in
its own change.

## Verification

- Triage core added to both credential-scan parametrize lists.
- Coverage is real, not nominal: **disabling the new scan makes
  `test_a_response_carrying_the_credential_fails_the_job` fail for this
  workflow specifically** (`StopIteration` — no step with `NEEDLES=`).
- `uv run pytest -q` → **1993 passed**
- `actionlint` clean; workflow parses

---


happyhuman:

PR #2645, opened 2026-09-17
chore: set ruff target-version to py311 to match requires-python

happyhuman:
`target-version = "py310"` sat two lines below `requires-python = ">=3.11"`,
and AGENTS.md sets the same 3.11 floor. Ruff was holding the codebase to a
version the repo does not support, withholding the 3.11-only rewrites it is
free to use.

## The fixes

All applied by `ruff --fix`; every one is an exact-equivalence rewrite:

| rule | change | why it's safe |
|---|---|---|
| `UP017` | `datetime.timezone.utc` → `datetime.UTC` | same object, 3.11 alias |
| `UP041` | `asyncio.TimeoutError` → `TimeoutError` | `asyncio.TimeoutError` **is** the builtin since 3.11 — a rename, not a behaviour change |
| `I001` | `tomllib` sorted into the stdlib group | it joined stdlib in 3.11 |

Checked `asyncio` is still used in all three files the `UP041` change touches,
so no import is left dangling.

## One deliberate suppression

`UP036` wanted to **delete** the `sys.version_info < (3, 11)` guard at the top
of `extract_env_vars.py`. That guard is not dead code — the file is a
standalone script a skill invokes directly, sometimes under a bare `python3`,
and the project's floor is a promise about the *repo*, not about the
interpreter this file is run with. Keeping it turns "wrong interpreter" into a
sentence that says so, rather than an `ImportError` on the `tomllib` line
immediately below.

## Scope — worth being explicit

**`.github/scripts/` and `.agents/` had to be included.** They were clean
before and gained **11 errors** under the new target. `python-format.yml` lints
every changed `.py` file, not just the recipe roots, so leaving them would have
turned any subsequent PR touching CI tooling red — **including the four
currently open** (#2641, #2642, #2643, #2644).

**The legacy `python/agents/` tree is deliberately untouched.** It goes 56 → 60
errors under the new target, but `python-format.yml` filters on
`core|contrib|skills/**/*.py` and `**/pyproject.toml`, so that tree is never
linted by CI and the delta is unreachable. It already carries 56 errors and 20
unformatted files; sweeping ~10 unrelated recipes to fix that belongs in its
own change, per AGENTS.md.

## Verification

```
ruff check .github/scripts .agents core contrib skills   ->  All checks passed!
ruff format --check <14 changed files>                   ->  all formatted
uv run pytest -q                                         ->  1991 passed
```

---


happyhuman:

